PR #45, opened 2024-01-03
feat: update OakQuizCheckbox feedback state

kimon-satan:
Implements new designs to visually represent the following states:

image

### to test

look at 

https://deploy-preview-45--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oakquizcheckbox--docs

- check all states rendered corrrectly

run 

```npm run test```

### DIFF
--- a/src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
+++ b/src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
@@ -138,8 +138,7 @@ const StyledFlexBox = styled(OakFlex)<StyledFlexBoxProps>`
 `;
 
 export type OakQuizCheckBoxProps = BaseCheckBoxProps & {
-  isFeedback?: boolean;
-  isCorrect?: boolean;
+  feedback?: "correct" | "incorrect" | null;
   image?: React.JSX.Element;
   innerRef?: React.RefObject<HTMLInputElement>;
 };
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx [hunk 1]
```diff
@@ -163,14 +186,25 @@ export const OakQuizCheckBox = (props: OakQuizCheckBoxProps) => {
...
   const inputCheckbox = (
     <StyledFlexBox
       $pa="inner-padding-l"
       $borderRadius={"border-radius-m2"}
       $borderColor={"border-primary"}
-      $background={disabled ? "bg-neutral-stronger" : "bg-primary"}
+      $background={backgroundColor}
       $flexGrow={1}
       onClick={handleContainerClick}
+      overlayBorderColor={isFeedback ? feedbackBorderColor : undefined}
```
carlmw, line 206: 🤔 If i'm following this correctly setting `isCorrect` won't do anything unless `isFeedback` is also true. The need for both could be helped with docs but its a little confusing.

Could this be simplified by replacing both booleans with a union?

Giving us something like:

```jsx
<>
  <OakQuizCheckbox feedback="correct" />
  <OakQuizCheckbox feedback="incorrect" />
</>
```

What do you think?

kimon-satan: yes that's right. I can see that's a bit cleaner but it's perhaps less so in the wider context. Each question has a feedback state boolean which can be easily passed on as a prop (see below). On the other hand I can now pass the feedback directly on instead. Radio buttons will also need a neutral feedback state. ... hmmm okay I'll do it :)

```
const isCorrect = questionState.feedback?.[index] === "correct";
        return (
          <OakQuizCheckBox
            key={`${questionUid}-answer${index}`}
            id={`${questionUid}-answer${index}`}
            value={answerText ? answerText.text : ""}
            isFeedback={isFeedbackMode}
            isCorrect={isCorrect}
            innerRef={innerRefs.current[index]}
            onChange={handleOnChange}
          />
        );
   ```

carlmw (approved):
🚀

rightsaidjames (approved):
Lookin' good! There may be more we can do about the semantics of the feedback for multiple choice questions in the future, but I think we need to see this in situ with real data before doing any further work on this.

PR #47, opened 2024-01-04
feat(PUPIL-86) Add `OakTextInput`

carlmw:
Adds `OakTextInput` which will be used initially to enable free text input for quizzes 

## TODO
- [ ] Update the corresponding search, tick and cross icons 

## Links

Figma 👉🏼  https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=3854-11120&mode=dev

To review in Storybook 👉🏼 https://deploy-preview-47--lively-meringue-8ebd43.netlify.app/?path=/docs/components-ui-oaktextinput--docs

## A/C

- [x] Focus state
- [x] Disabled state
- [x] Read only state
- [x] Valid
- [x] Invalid
- [x] Start and end enhancer icons

### DIFF
--- a/src/components/base/InternalTextInput/InternalTextInput.stories.tsx
+++ b/src/components/base/InternalTextInput/InternalTextInput.stories.tsx
@@ -3,30 +3,16 @@ import { Meta, StoryObj } from "@storybook/react";
 
 import { InternalTextInput } from "./InternalTextInput";
 
-/**
- *
- * An unstyled input to be used as a basis for UI input components.
- * The following callbacks are available for tracking focus events:
- *
- *  ### onFocus
- * `(e: FocusEvent<HTMLInputElement>) => void;`
- *  ### onBlur
- * `(e: FocusEvent<HTMLInputElement>) => void;`
- *  ### onInitialFocus
- * `(e: FocusEvent<HTMLInputElement>) => void;`<br>
- *  occurs only when the input is focused for the first time
- *
- */
+import { sizeArgTypes } from "@/storybook-helpers/sizeStyleHelpers";
+import { spacingArgTypes } from "@/storybook-helpers/spacingStyleHelpers";
 
 const meta: Meta<typeof InternalTextInput> = {
   component: InternalTextInput,
   tags: ["autodocs"],
   title: "components/base/InternalTextInput",
-  argTypes: {},
-  parameters: {
-    controls: {
-      include: [],
-    },
+  argTypes: {
+    ...sizeArgTypes,
+    ...spacingArgTypes,
   },
 };
 export default meta;
@@ -37,8 +23,4 @@ export const Default: Story = {
   render: (args) => (
     <InternalTextInput {...args} placeholder="placeholder text" type="text" />
   ),
-  args: {},
-  parameters: {
-    controls: { include: [] },
-  },
 };
--- a/src/components/base/InternalTextInput/InternalTextInput.stories.tsx
+++ b/src/components/base/InternalTextInput/InternalTextInput.stories.tsx
@@ -3,16 +3,14 @@ import { Meta, StoryObj } from "@storybook/react";
 
 import { InternalTextInput } from "./InternalTextInput";
 
-import { sizeArgTypes } from "@/storybook-helpers/sizeStyleHelpers";
-import { spacingArgTypes } from "@/storybook-helpers/spacingStyleHelpers";
-
 const meta: Meta<typeof InternalTextInput> = {
   component: InternalTextInput,
   tags: ["autodocs"],
   title: "components/base/InternalTextInput",
-  argTypes: {
-    ...sizeArgTypes,
-    ...spacingArgTypes,
+  parameters: {
+    controls: {
+      include: [],
+    },
   },
 };
 export default meta;
--- a/src/components/base/InternalTextInput/InternalTextInput.tsx
+++ b/src/components/base/InternalTextInput/InternalTextInput.tsx
@@ -33,6 +33,7 @@ const StyledInput = styled.input`
   font-family: inherit;
   background: transparent;
   outline: none;
+  color: inherit;
 
   @media (max-width: ${getBreakpoint("small")}px) {
     /* iOS zooms in on inputs with font sizes <16px on mobile */
@@ -68,7 +69,7 @@ export type InternalTextInputProps = StyledInputProps & {
  * An unstyled input to be used as a basis for UI input components.
  * Supports all the props of a regular `HTMLInputElement`
  *
- * The CSS `outline` is disabled so a focus ring must be applied by the consuming component.
+ * 🚨 The CSS `outline` is disabled so a focus ring must be applied by the consuming component.
  *
  * The following callbacks are available for tracking focus events:
  *
--- a/src/components/base/InternalTextInput/InternalTextInput.tsx
+++ b/src/components/base/InternalTextInput/InternalTextInput.tsx
@@ -8,11 +8,16 @@ import styled from "styled-components";
 
 import { parseColor } from "@/styles/helpers/parseColor";
 import { getBreakpoint } from "@/styles/utils/responsiveStyle";
+import { SpacingStyleProps, spacingStyle } from "@/styles/utils/spacingStyle";
+import { SizeStyleProps, sizeStyle } from "@/styles/utils/sizeStyle";
 
 type StyledInputProps = Omit<
   DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>,
   "ref"
->;
+> &
+  SpacingStyleProps &
+  SizeStyleProps;
+
 /**
  * Using `appearance none !important;` here because many style resets will set this
  * value to textfield, causing some browsers to implement undesirable styles.
@@ -26,6 +31,8 @@ const StyledInput = styled.input`
   border-color: transparent;
   box-shadow: none;
   font-family: inherit;
+  background: transparent;
+  outline: none;
 
   @media (max-width: ${getBreakpoint("small")}px) {
     /* iOS zooms in on inputs with font sizes <16px on mobile */
@@ -43,16 +50,39 @@ const StyledInput = styled.input`
   ::-webkit-search-results-decoration {
     appearance: none;
   }
+
+  ${spacingStyle}
+  ${sizeStyle}
 `;
 
 export type InternalTextInputProps = StyledInputProps & {
-  onFocus?: (e: FocusEvent<HTMLInputElement>) => void;
-  onBlur?: (e: FocusEvent<HTMLInputElement>) => void;
+  placeholder?: string;
+  /**
+   * Fired only when the input is focused for the first time
+   */
   onInitialFocus?: (e: FocusEvent<HTMLInputElement>) => void;
 };
 
+/**
+ *
+ * An unstyled input to be used as a basis for UI input components.
+ * Supports all the props of a regular `HTMLInputElement`
+ *
+ * The CSS `outline` is disabled so a focus ring must be applied by the consuming component.
+ *
+ * The following callbacks are available for tracking focus events:
+ *
+ *  ### onFocus
+ * `(e: FocusEvent<HTMLInputElement>) => void;`
+ *  ### onBlur
+ * `(e: FocusEvent<HTMLInputElement>) => void;`
+ *  ### onInitialFocus
+ * `(e: FocusEvent<HTMLInputElement>) => void;`<br>
+ *  occurs only when the input is focused for the first time
+ *
+ */
 export const InternalTextInput = (props: InternalTextInputProps) => {
-  const { onInitialFocus, onBlur, onFocus, ...rest } = props;
+  const { onInitialFocus, onFocus, ...rest } = props;
 
   const hadInitialFocused = useRef(false);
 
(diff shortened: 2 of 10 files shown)


src/components/ui/OakTextInput/OakTextInput.tsx
```diff
@@ -0,0 +1,196 @@
...
+  $invalidIconColor = "border-error",
+  startEnhancerIconName,
+  endEnhancerIconName,
+  ...props
+}: OakTextInputProps) => {
+  let borderColor: OakCombinedColorToken;
+  let iconColor: OakCombinedColorToken;
+  let readOnlyBorderColor: OakCombinedColorToken;
+
+  switch (validity) {
```
carlmw, line 132: I'm in two minds as to whether the validation styles warrant another flavour of input component.

On the one hand it is "logic", but validation seems intrinsic to a form field so perhaps it belongs at this level since the colours can still be set through style props.

My take away from what i've seen in the library so far is that the intention is for components to become more opinionated through each layer of composition.

So perhaps this does belong here, but `OakQuizInput` is where we apply an end enhancer icon of tick/cross and validity `'valid' | 'invalid'`when the feedback is `'correct' | 'incorrect'`



`base/InternalTextInput` => unstyled, fully transparent but with the ability to set dimensions and spacing. Intended for consumption by a bonafide component ready for use
`ui/OakTextInput` => some opinionated branded styling with strong defaults and baked in states. Allows any colour to be changed through styling props
`integrated/OakQuizInput` =>  very opinionated, pretty much single use.

🤔 thinking about it, does `OakQuizInput` even have a place in a component library. Is it so specific that it actually belongs in the app?

src/components/ui/OakTextInput/OakTextInput.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  background: ${(props) => parseColor(props.$background)};
+
+  @media (hover: hover) {
+    &:hover:not(:focus-within, :has(input:read-only)) {
+      background: ${(props) => parseColor(props.$hoverBackground)};
+    }
+  }
+
+  &:has(input:read-only) {
```
carlmw, line 90: I've seen use of `:has` elsewhere, are we okay targeting only evergreen browsers?

rightsaidjames: Looking at https://caniuse.com/css-has versus https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2022, I think it's worth exploring potential polyfills* or alternative solutions here.

*we'd need to targeting Firefox `> v103` (desktop) and Safari `> 15.4` (desktop and mobile)

rightsaidjames: ... but it depends if the native fallback is functional and accessible (or not)

src/components/ui/OakTextInput/OakTextInput.tsx
rightsaidjames: I don't have any opinions on your last paragraph (all I care about here is that the base components be independently testable and well-documented), but I agree with the assumptions in the rest of your comment

kimon-satan: On the last para there's something of a fuzzy line but the rule I'm applying is that if the component is "dumb" ie. still doesn't require knowledge of context then it can live in this library. The ultimate aim here is that we have components which can be used in multiple apps. A quiz checkbox could certainly be one of those.

src/components/ui/OakTextInput/OakTextInput.tsx
kimon-satan: given the styling of requirements of the quiz checkbox an alternative solution to :has would be a pretty heavy rewrite and we're already behind. Polyfill sounds like a a good idea to me. Looking at your above comment James it seems like the main issue would be safari iOS. I'd still put this on the backlog.

src/components/ui/OakTextInput/OakTextInput.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+export const Invalid: Story = {
+  render: () => <OakTextInput validity="invalid" value="A fine text value" />,
+};
+
+export const ValidWithStartEnhancer: Story = {
+  render: () => (
+    <OakTextInput
+      validity="valid"
+      value="A fine text value"
+      startEnhancerIconName="tick"
```
rightsaidjames, line 81: Alt text for icons should generally convey meaning not appearance, so this should be 'correct', I think?

src/components/ui/OakTextInput/OakTextInput.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    />
+  ),
+};
+
+export const InvalidWithEndEnhancer: Story = {
+  render: () => (
+    <OakTextInput
+      validity="invalid"
+      value="A fine text value"
+      endEnhancerIconName="cross"
```
rightsaidjames, line 91: As above, but `incorrect`

src/components/ui/OakTextInput/OakTextInput.stories.tsx
rightsaidjames: Actually this is just the icon name, isn't it? In which case my beef is with `OakIcon`, not this file.

src/components/base/InternalTextInput/InternalTextInput.stories.tsx
```diff
@@ -3,30 +3,16 @@ import { Meta, StoryObj } from "@storybook/react";
 
 import { InternalTextInput } from "./InternalTextInput";
 
-/**
- *
- * An unstyled input to be used as a basis for UI input components.
```
kimon-satan, line 8: I'd prefer to keep this documentation and amend if necessary.

src/components/base/InternalTextInput/InternalTextInput.stories.tsx
```diff
@@ -3,30 +3,16 @@ import { Meta, StoryObj } from "@storybook/react";
...
+import { spacingArgTypes } from "@/storybook-helpers/spacingStyleHelpers";
 
 const meta: Meta<typeof InternalTextInput> = {
   component: InternalTextInput,
   tags: ["autodocs"],
   title: "components/base/InternalTextInput",
-  argTypes: {},
-  parameters: {
-    controls: {
-      include: [],
```
kimon-satan, line 28: removing this field populates the story with all possible props. The designers complained that these made the stories unnavigable for them. We therefore adopt an approach of putting a curated list of props in the story - namely those props which can be meaningfully manipulated and are specific to that component.

src/components/base/InternalTextInput/InternalTextInput.stories.tsx
kimon-satan: Whoops just saw you added it elsewhere - disregard

src/components/base/InternalTextInput/InternalTextInput.tsx
```diff
@@ -43,33 +51,52 @@ const StyledInput = styled.input`
...
   const handleOnFocus = (e: FocusEvent<HTMLInputElement>) => {
     if (!hadInitialFocused.current && props.onInitialFocus) {
       props.onInitialFocus(e);
       hadInitialFocused.current = true;
-    } else if (props.onFocus) {
-      props.onFocus(e);
     }
-  };
 
-  const handleOnBlur = (e: FocusEvent<HTMLInputElement>) => {
```
kimon-satan, line 68: why remove the blur event handler ?

src/components/ui/OakTextInput/OakTextInput.stories.tsx
kimon-satan: OakIcon defaults to the Icon name but you can override to whatever you want.

src/components/ui/OakTextInput/OakTextInput.tsx
```diff
@@ -0,0 +1,196 @@
...
+    color: ${(props) => parseColor(props.$disabledColor)};
+  }
+`;
+
+/**
+ * Default input which can be extended to create specialised inputs.
+ */
+export const OakTextInput = ({
+  type = "text",
+  $borderColor = "border-primary",
```
kimon-satan, line 106: if these aren't actually used by styled components then they shouldn't be prefixed with the $ . The reason being that should we later decide to export this as styled-component to allow overrides of styles the $ prefixed props will be consumed before hitting the TSX

src/components/ui/OakTextInput/OakTextInput.tsx
```diff
@@ -0,0 +1,196 @@
...
+      $readOnlyBorderColor={readOnlyBorderColor}
+      $disabledColor={$disabledColor}
+      $readOnlyColor={$readOnlyColor}
+      $color={$color}
+      $alignItems="center"
+      $position="relative"
+      $gap="space-between-s"
+      $ph="inner-padding-l"
+      onClick={(event) => {
+        event.currentTarget.querySelector("input")?.focus();
```
kimon-satan, line 171: Should we pass this event up as a callback ?

src/components/ui/OakTextInput/OakTextInput.tsx
```diff
@@ -0,0 +1,196 @@
...
+  $color = "text-primary",
+  $disabledColor = "text-disabled",
+  $readOnlyColor = "text-subdued",
+  validity,
+  $iconColor = "icon-inverted",
+  $validBorderColor = "border-success",
+  $invalidBorderColor = "border-error",
+  $validIconColor = "icon-success",
+  $invalidIconColor = "border-error",
+  startEnhancerIconName,
```
kimon-satan, line 124: Could we follow the same interface as in InternalButton ?

kimon-satan (changes requested):
Sorry lots of comment here. Not sure I'm right about everything so lets discuss and come up with some agreed principles which we can follow through in other components

src/components/base/InternalTextInput/InternalTextInput.tsx
carlmw: The code was redundant since the prop is forwarded to the underlying input in the rest arg

src/components/ui/OakTextInput/OakTextInput.tsx
carlmw: 🤔 thinking about this more I wonder if the more appropriate thing to do here is to also trigger `click` on the input.

The intent was for the div to proxy the input. If I bound an `onClick` to `OakInput` and `currentTarget` was the wrapper `div` I wouldn't have immediate access to the value of the `input` which might be unexpected.

As it stands if I pass `onClick` to `OakInput` it will be bound to the underlying `<input />`  but won't be called if I click the padding of the `div`.

```suggestion
        const domInput = event.currentTarget.querySelector("input");
        input?.focus();
        input?.click();
```

src/components/base/InternalTextInput/InternalTextInput.stories.tsx
```diff
@@ -37,8 +21,4 @@ export const Default: Story = {
   render: (args) => (
     <InternalTextInput {...args} placeholder="placeholder text" type="text" />
   ),
-  args: {},
-  parameters: {
```
carlmw, line 41: These were not required since the `include` is first taken from the `Meta`

src/components/ui/OakTextInput/OakTextInput.tsx
carlmw: Okay, I think is now moot for `OakInput` at least. The only a11y issue I think it introduced was the lack of a focus ring in unsupported browsers. However I was only using `:has` to disable the focus styles when the input was `readOnly` — in retrospect this was incorrect. As per https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions a `readOnly` input should still receive focus.

src/components/ui/OakTextInput/OakTextInput.stories.tsx
carlmw: 🤔 in this case I think the icons are presentational since they're not strictly content. It's only a quirk of implementation that its not added as a background image or another CSS decoration.

It's probably outside the scope of this component (and more the job of a specific input component geared toward quizzes) but would something like `aria-invalid` with `aria-describedBy` be appropriate? https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA21.html

I'll be moving onto that next I suspect so I might ask for feedback when that comes up.

kimon-satan (approved):
Really great ! Thanks for implementing all of those.

src/components/ui/OakTextInput/OakTextInput.stories.tsx
rightsaidjames: @carlmw In this context it's decorative because you're merely demonstrating the capabilities of the component. However in-situ we need to consider what the meaning/purposes of these icons are - if it's displaying a tick or cross (after the user has submitted their answer to be marked) then the icon is not merely presentational/decorative.

I would argue that `aria-invalid` is the wrong choice here for post-submission feedback, because the input itself was valid, but the answer the user gave was incorrect as per the marking scheme. And as far as I know there's no ARIA property called `aria-valid` or similar.

That being said, I'm happy to park this for now and revisit it when the `OakQuizTextInput` component (or whatever) is created.

PR #51, opened 2024-01-08
feat(PUPIL-214) `OakQuizInput` component

carlmw:
* Adds the `not-allowed` cursor to inputs when disabled
* Adds an `iconAlt` prop to `OakTextInput` so that we can deliver some context to AT with the icon
* Adds `OakQuizTextInput` which is a specialised form of `OakTextInput` configured with feedback states for quiizzes.

## TODO
- [ ] Add 👇🏼 new icons when I have access

search
warning

Builds upon the base `OakTextInput` as in Figma here 👇🏼 
https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=3854-11120&mode=dev

Can be found in Storybook here 👇🏼 
https://deploy-preview-51--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oakquiztextinput--docs

### DIFF
--- a/src/components/integrated/OakQuizTextInput/OakQuizTextInput.tsx
+++ b/src/components/integrated/OakQuizTextInput/OakQuizTextInput.tsx
@@ -1,0 +1,48 @@
+import React from "react";
+
+import { OakTextInput, OakTextInputProps } from "@/components/ui";
+
+type OakQuizTextInputProps = Omit<
+  OakTextInputProps,
+  "validity" | "iconName" | "iconAlt" | "isTrailingIcon"
+> & {
+  /**
+   * Alters the appearance of the input to indicate whether or not a correct answer was given.
+   * Also sets the input to read-only.
+   */
+  feedback?: "correct" | "incorrect" | null;
+};
+
+export const OakQuizTextInput = ({
+  feedback,
+  readOnly,
+  ...props
+}: OakQuizTextInputProps) => {
+  let validity: OakTextInputProps["validity"];
+  let iconName: OakTextInputProps["iconName"];
+  let iconAlt: OakTextInputProps["iconAlt"] = undefined;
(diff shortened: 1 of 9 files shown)


src/components/integrated/OakQuizTextInput/OakQuizTextInput.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
```
carlmw, line 1: @kimon-satan I'm conflicted as to whether this belongs in `ui`, `integrated` or maybe even `base` within the current structure.

Going by atomic design this would probably be classed as an "atom". But i'm given pause by this line in the integrated component docs:

> Applicable to a limited range of contexts which might occur in multiple apps 

This is a pretty specialised flavour of input, but it isn't overly complex and I wouldn't put it in the "molecule" box of atomic design.

To go with the Lego analogy, I see `OakTextInput` as a "brick" but `OakQuizTextInput` is some flavour of ["brick modified"](https://www.bricklink.com/v2/catalog/catalogitem.page?P=2877#T=C) — still a brick, but with a more specific purpose.

If I gave this component responsibility for the field label and brought in `OakLabel` it would probably more comfortably fit into the "molecule" camp.

Interested in your take 🙇🏼

kimon-satan: Yeah I can see the problem.  Lets consider the folder rules like a ratchet - start at the bottom of the hierarchy if a rule fails move up a level. In this case it fails the context rules in base and ui so it should belong in integrated even though it doesn't pass all the tests there . Alternatively just view it as this ;)

image

kimon-satan (approved):

PR #54, opened 2024-01-09
chore(deps): next 14 update

johnrobeds:
Alignment to OWA packages and node 20 to allow for Next 14 testing


jimCresswell (approved):

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@oak-academy/oak-components",
-  "version": "0.0.19",
+  "version": "0.0.20",
```
kimon-satan: This will automatically increment on publish so shouldn't be changed here.

kimon-satan (changes requested):
Cheers for doing this. Before merging the build needs to be tested against a version of OWA which uses next 14 before publishing or else it will block pupil squad's work. This can be done using yalc (see publish:local in package.json)

jimCresswell:
> Cheers for doing this. Before merging the build needs to be tested against a version of OWA which uses next 14 before publishing or else it will block pupil squad's work. This can be done using yalc (see publish:local in package.json)

@kimon-satan @johnrobeds are we satisfied this has been tested now?

kimon-satan (approved):

PR #59, opened 2024-01-11
feat(PUPIL-232): add `OakRoundIcon` component

carlmw:
Adds a variant of `OakIcon` with a rounded background and the ability to change colors.

Figma 👉🏼 https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=3923-3041&mode=dev

Storybook 👉🏼  https://deploy-preview-59--lively-meringue-8ebd43.netlify.app/?path=/docs/components-ui-oakroundicon--docs

deploy-preview-59--lively-meringue-8ebd43 netlify app__path=_docs_components-ui-oakroundicon--docs


src/components/ui/OakRoundIcon/OakRoundIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export const OakRoundIcon = ({
+  $background = "bg-decorative5-main",
+  $borderRadius = "border-radius-circle",
+  $width = "all-spacing-9",
+  $height = "all-spacing-9",
+  $pa = "inner-padding-ssx",
+  ...rest
+}: OakRoundIconProps) => {
+  const icon = rest.iconName ? (
```
carlmw, line 33: 🤔 I was thinking that there would be cases where the icon itself would not belong in the icon library (I.e a one shot/super bespoke SVG)

In that case it would be useful to be able to point to the `src` of the asset instead of an `iconName` 

I might be a little off course on that assumption in which case we can remove this.

src/components/ui/OakRoundIcon/OakRoundIcon.stories.tsx
```diff
@@ -0,0 +1,95 @@
+import React from "react";
+import { Meta, StoryObj } from "@storybook/react";
+
+// @ts-expect-error:  this is an svg
+import lightbulbOutlineIconSvg from "../../../../assets/light-bulb-outline.svg";
+// @ts-expect-error:  this is an svg
+import lightbulbWhiteOutlineSvg from "../../../../assets/light-bulb-with-white-outline.svg";
```
carlmw, line 7: These icons have some very particular colour choices which don't lend themselves easily to a monochromatic colour filter. 

I was thinking (at least for now) we could use two separate assets (one with the black stroke and another with the white and yellow fill). Going forward it would be great to be able to apply CSS to icons to target `fill` and `stroke` but that would only be possible with an inline SVG.

Thoughts are very welcome. We could also push back on the multi-colour icon for now?

src/components/ui/OakRoundIcon/OakRoundIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+  src?: undefined;
+};
+
+export type OakRoundIconProps = IconProps | ImageProps;
+
+/**
+ * A wrapper around `OakIcon` which applies a rounded background.
+ * Supports either an icon name or an image src
+ */
+export const OakRoundIcon = ({
```
carlmw, line 25: I thought `OakRoundIcon` was more descriptive than `OakContainedIcon`, happy to be challenged though!

PR #61, opened 2024-01-12
feat(PUPIL-230) Add `OakQuizBottom`

carlmw:
Supercedes https://github.com/oaknational/oak-components/pull/60


src/styles/utils/borderStyle.ts
```diff
@@ -64,6 +70,58 @@ const borderRadius = css<{ $borderRadius?: BorderRadiusProps }>`
     parseBorderRadius,
   )}
 `;
+const borderRadiusTopLeft = css<BorderStyleProps>`
```
carlmw, line 73: I wanted to be able to control the border-radius for each corner in the tooltip. I could have added this with styled-components but I felt that this possibly belonged here.

src/components/integrated/OakQuizHint/OakQuizHint.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Presents a button which will show a hint when clicked
+ */
+export const OakQuizHint = ({ hint }: OakQuizHintProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const handleClick: MouseEventHandler = () => {
+    setIsOpen(!isOpen);
+  };
+
+  return (
+    <OakTooltip tooltip={hint} isOpen={isOpen}>
```
carlmw, line 22: Should this be a tooltip or a modal that looks like a tooltip? 🦆 

It isn't triggered on hover, only on click so perhaps it should be viewed as `role="modal"`  for AT?

src/components/base/OakBox/__snapshots__/OakBox.test.tsx.snap
```diff
@@ -2,6 +2,6 @@
 
 exports[`Component OakBox matches snapshot 1`] = `
 <div
-  className="sc-aXZVg goWQoj"
+  className="sc-aXZVg dPEvkO"
 />
```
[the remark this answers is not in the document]
carlmw: It's where I added those border radii styling props to `OakBox`. Class names are generated from (among other things) a hash of the style strings — so any changes to styles will change class names.

PR #66, opened 2024-01-15
feat(PUPIL-256): add `OakBackLink` component

carlmw:
a button or link to navigate back to the previous page in the user's journey. This component will initially be used in the top nav for the pupil lesson flow.

Storybook 👉🏼  https://deploy-preview-66--lively-meringue-8ebd43.netlify.app/?path=/docs/components-ui-oakbacklink--docs

Screenshot 2024-01-16 at 11 28 28


carlmw:
> I can see you've just followed the figma but I'm wondering about the inconsistency in design here. Why are the hover and focus states different to the tertiary button and other buttons ?

I think the tertiary button needs revisiting for sure.

PR #67, opened 2024-01-16
feat(PUPIL-229): add `OakLessonTopNav` component

carlmw:
This will render the navigation bar at the top of a lesson.

It includes slots for the back link and counter. These are included to keep the component dumb and leave it up to the host to decide how to hook up the back link and counter to the lesson state.

Figma 👉🏼  https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=4604-14267&mode=dev

Storybook 👉🏼 https://deploy-preview-67--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oaklessontopnav--docs 

A `mobileSummary` prop is included to allow the host to pass in alternate content when at the mobile breakpoint since the counter is not displayed.

**Mobile**
Screenshot 2024-01-16 at 15 37 06

**Above**
Screenshot 2024-01-16 at 15 37 11

### DIFF
--- a/src/components/base/OakIcon/OakIcon.tsx
+++ b/src/components/base/OakIcon/OakIcon.tsx
@@ -52,9 +52,7 @@ export const oakIconNames = [
   "warning",
   "lightbulb",
   "lightbulb-yellow",
-  "signed-video-4",
-  "video-3",
-  "quiz-6",
+  "intro",
 ] as const;
 
 export type OakIconName = (typeof oakIconNames)[number];
--- a/src/components/base/OakIcon/OakIcon.tsx
+++ b/src/components/base/OakIcon/OakIcon.tsx
@@ -52,6 +52,9 @@ export const oakIconNames = [
   "warning",
   "lightbulb",
   "lightbulb-yellow",
+  "signed-video-4",
+  "video-3",
+  "quiz-6",
 ] as const;
 
 export type OakIconName = (typeof oakIconNames)[number];
--- a/src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
+++ b/src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
@@ -1,0 +1,100 @@
+import React, { ReactNode } from "react";
+import styled from "styled-components";
+
+import {
+  OakBox,
+  OakFlex,
+  OakHeading,
+  OakIconProps,
+  OakSpan,
+} from "@/components/base";
+import { OakRoundIcon } from "@/components/ui";
+import { getBreakpoint } from "@/styles/utils/responsiveStyle";
+
+type LessonSectionName = "intro" | "starter-quiz" | "video" | "exit-quiz";
+
+const StyledMobileSummary = styled(OakSpan)`
+  @media (min-width: ${getBreakpoint("small")}px) {
+    display: none;
+  }
+`;
+
+export type OakLessonTopNavProps = {
+  lessonSectionName: LessonSectionName;
+  /**
+   * Slot to render `OakBackLink` or similar
+   */
+  backLinkSlot: ReactNode;
+  heading: ReactNode;
+  /**
+   * Displayed at the mobile breakpoint where the counter is not rendered.
+   * Provides alternative content for the counter/progress in the lesson.
+   */
+  mobileSummary: ReactNode;
+  /**
+   * Slot to render `OakQuizCounter` or similar
+   */
+  counterSlot?: ReactNode;
+};
+
+/**
+ * Controls for navigating back and displaying progress in a lesson
+ */
+export const OakLessonTopNav = ({
+  lessonSectionName,
+  backLinkSlot,
+  counterSlot,
+  heading,
+  mobileSummary,
+}: OakLessonTopNavProps) => {
+  return (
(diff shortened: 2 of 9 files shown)


src/components/base/OakIcon/OakIcon.tsx
```diff
@@ -52,6 +52,9 @@ export const oakIconNames = [
   "warning",
   "lightbulb",
   "lightbulb-yellow",
+  "signed-video-4",
+  "video-3",
+  "quiz-6",
```
carlmw, line 57: These match the names of the assets in Figma, but i'm feeling that this is going to become a quagmire. 

Feels like we need a better reference in Figmaland to map these properly.

src/components/integrated/index.ts
```diff
@@ -2,3 +2,7 @@ export * from "./OakQuizCheckBox";
 export * from "./OakQuizTextInput";
 export * from "./OakQuizRadioButton";
 export * from "./OakQuizHint";
+export * from "./OakQuizBottomNav";
+export * from "./OakLessonTopNav";
+export * from "./OakQuizFeedback";
+export * from "./OakQuizCounter";
```
carlmw, line 8: Every... time... 😆 🤦🏼 

Some sort of lint check might work, or it could be solved with a checklist and good ol' fashioned thoroughness.

src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Slot to render `OakQuizCounter` or similar
+   */
+  counterSlot?: ReactNode;
+};
+
+/**
+ * Controls for navigating back and displaying progress in a lesson
+ */
+export const OakLessonTopNav = ({
```
carlmw, line 43: I'm going to rename `OakQuizBottomNav` to `OakLessonBottomNav` in my next ticket (which is to add the lesson layout). In hindsight it didn't occur to me that the bottom nav actually applied to the whole lesson context.

I'm thinking something like:

```jsx
<OakLessonLayout topNavSlot={<OakLessonTopNav />} bottomNavSlot={<OakLessonBottomNav />}>
  current lesson content goes here
</OakLessonLayout>
```

That way in OWA `OakLessonTopNav` and `OakLessonBottomNav` can be wrapped in something that gives them access to the lesson context before being passed into their slots.

src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Slot to render `OakBackLink` or similar
+   */
+  backLinkSlot: ReactNode;
+  heading: ReactNode;
+  /**
+   * Displayed at the mobile breakpoint where the counter is not rendered.
+   * Provides alternative content for the counter/progress in the lesson.
+   */
+  mobileSummary: ReactNode;
```
carlmw, line 33: This leaves me feeling a bit icky but the design calls for the counter to be hidden on mobile and for some extra text to be displayed below the heading instead. 

Screenshot 2024-01-16 at 15 52 47
Screenshot 2024-01-16 at 15 52 43

I'm hiding the counter completely with `display: none` on mobile and hiding the `mobileSummary` with `display: none` when above mobile size. It doesn't seem to make sense to display both to AT at the same time so this seems like a sensible approach to me.

Might be a question for @rightsaidjames but can anyone see any pitfalls with this approach?

src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
```diff
@@ -0,0 +1,100 @@
...
+}: OakLessonTopNavProps) => {
+  return (
+    <OakFlex $gap="space-between-m" $alignItems="center">
+      {backLinkSlot}
+      <OakFlex $flexGrow="none">
+        <OakRoundIcon
+          {...pickSectionIcon(lessonSectionName)}
+          $width="all-spacing-8"
+          $height="all-spacing-8"
+        />
```
carlmw, line 58: Foregoing the wibbly wobbly background for now and going with a plain round icon.

src/components/integrated/OakLessonTopNav/OakLessonTopNav.tsx
rightsaidjames: This seems like the correct approach to me - AT should (and does!) respect `display: none`* and media queries, so giving them an equivalent experience rather than a dedicated one is clearly the right solution in this scenario.

*_and `block`, `inline`, `inline-block` etc._

src/components/base/OakIcon/OakIcon.tsx
kimon-satan: Could you speak to Mike about these ?

messmike: @carlmw the names of these svg icons have now been updated in the Figma design kit to the following:
- intro
- video
- quiz

I'll check through other svgs to make sure naming makes sense.

kimon-satan (approved):

PR #70, opened 2024-01-18
feat(PUPIL-257) `OakLessonNavItem`

carlmw:
Enables navigation to the given section of a lesson as well as displaying current progress in that section.

Figma 👉🏼  https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=4604-17936&mode=dev

Storybook 👉🏼 https://deploy-preview-70--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oaklessonnavitem--docs

**Tablet >=**
Screenshot 2024-01-18 at 17 23 36

**Mobile**
Screenshot 2024-01-18 at 17 27 46

### DIFF
--- a/src/components/integrated/OakLessonNavItem/OakLessonNavItem.stories.tsx
+++ b/src/components/integrated/OakLessonNavItem/OakLessonNavItem.stories.tsx
@@ -49,7 +49,7 @@ const meta: Meta<typeof OakLessonNavItem> = {
       include: [
         "progress",
         "lessonSectionName",
-        "isDisabled",
+        "disabled",
         "grade",
         "numQuestions",
         "videoLength",
--- a/src/components/integrated/OakLessonNavItem/OakLessonNavItem.stories.tsx
+++ b/src/components/integrated/OakLessonNavItem/OakLessonNavItem.stories.tsx
@@ -1,0 +1,90 @@
+import React from "react";
+import { Meta, StoryObj } from "@storybook/react";
+
+import { OakLessonNavItem } from "./OakLessonNavItem";
+
+import { OakFlex } from "@/components/base";
+
+const meta: Meta<typeof OakLessonNavItem> = {
+  component: OakLessonNavItem,
+  tags: ["autodocs"],
+  title: "components/integrated/OakLessonNavItem",
+  argTypes: {
+    label: {
+      control: {
+        type: "text",
+      },
+    },
+    summary: {
+      control: {
+        type: "text",
+      },
+    },
+  },
+  args: {
+    lessonSectionName: "intro",
+    progress: "not-started",
+    label: "Intro",
+    summary: "Get ready",
+    href: "#",
+  },
+  parameters: {
+    controls: {
+      include: [
+        "label",
+        "summary",
+        "progress",
+        "lessonSectionName",
+        "isDisabled",
+      ],
+    },
+  },
+};
+export default meta;
+
+type Story = StoryObj<typeof OakLessonNavItem>;
+
+export const Default: Story = {
+  render: (args) => <OakLessonNavItem {...args} />,
+};
+
+export const NotStarted: Story = {
+  render: () => (
+    <>
+      <OakLessonNavItem
+        lessonSectionName="intro"
+        progress="not-started"
+        label="Intro"
+      />
+      <OakLessonNavItem
+        lessonSectionName="starter-quiz"
+        progress="not-started"
+        label="Starter quiz"
+      />
+      <OakLessonNavItem
+        lessonSectionName="video"
+        progress="not-started"
+        label="Video"
+      />
+      <OakLessonNavItem
+        lessonSectionName="exit-quiz"
+        progress="not-started"
+        label="Exit quiz"
+      />
+    </>
+  ),
+  decorators: [
+    (Story) => {
+      return (
+        <OakFlex $flexDirection="column" $gap="space-between-m">
+          {Story()}
+        </OakFlex>
+      );
+    },
+  ],
+  parameters: {
+    controls: {
+      disable: true,
+    },
+  },
+};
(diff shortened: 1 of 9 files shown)


src/components/integrated/OakLessonNavItem/OakLessonNavItem.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+  tags: ["autodocs"],
+  title: "components/integrated/OakLessonNavItem",
+  argTypes: {
+    summary: {
+      control: {
+        type: "text",
+      },
+    },
+    videoLength: {
+      if: { arg: "lessonSectionName", eq: "video" },
```
carlmw, line 19: I wanted to conditionally show the `grade` and `numQuestion` controls only for `exit-quiz` and `starter-quiz` but this condition can only contain one value 🤬

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
```diff
@@ -0,0 +1,304 @@
...
+      return (
+        <>
+          <OakBox $display={["block", "none"]}>Completed</OakBox>
+          <OakBox $display={["none", "block"]}>
+            {props.grade}/{props.numQuestions} correct
+          </OakBox>
+        </>
+      );
+  }
+}
```
carlmw, line 304: I was in two minds as to whether the logic to pick the content belonged in here. I was already doing this to pick the icon so this didn't feel like a huge leap. My aim was to have something that could be readily hooked up to the relevant state in OWA.

If it wasn't here we'd have to lift all this into OWA, perhaps that is the right thing to do? Am I straying from the path?

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
```diff
@@ -0,0 +1,304 @@
...
+        >
+          {pickLabelForSection(lessonSectionName)}
+        </StyledLabel>
+        <OakBox $font={["body-2", "body-1"]}>
+          {pickSummaryForProgress(props)}
+        </OakBox>
+      </OakBox>
+      {renderQuestionCounter(props)}
+      <StyledRoundIcon iconName="chevron-right" $isDisabled={isDisabled} />
+    </StyledLessonNavItem>
```
carlmw, line 167: This could easily be extracted into a more general `OakNavItem` in a follow-up PR if this is something that makes sense as a reusable piece.

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
```diff
@@ -0,0 +1,304 @@
...
+    css`
+      cursor: pointer;
+
+      &:hover,
+      &:active {
+        ${StyledLabel} {
+          text-decoration: underline;
+        }
+
+        ${StyledRoundIcon} {
```
carlmw, line 104: This breaks encapsulation of component styles in `OakRoundIcon` and I loath it. My only other option was to write more JavaScript and use pointer events to flip some props. I'd loath that even more.

rightsaidjames (approved):
:chefs-kiss:

kimon-satan (changes requested):
It still seems to be clickable in its disabled state ?

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
carlmw: Not sure what you mean by inherit here?

carlmw:
> It still seems to be clickable in its disabled state ?

that is correct, because the component could also be a `<button>` or a `<a>`

I left this comment on the `isDisabled` prop as a guide
https://github.com/oaknational/oak-components/blob/2c32329b4745f3e364c89cee583563b9fb65f9ac/src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx#L22-L28

I was intending to leave it up to the consumer to either a. apply the `disabled` prop in the case of a `button` or b. replace the `a` with a `span` in the case of a link (since an `<a>` cannot be disabled).

Coming back to this on a Monday I think this is probably expecting too much of the consumer and will probably lead to avoidable misuse. I reckon I can bake this decision into the component and make it a `span` whenever it should be disabled.

carlmw:
@kimon-satan I added the logic to make the element a `span` when it's disabled. In hindsight it is the better API and would match my expectations as a consumer. Always good to remove a foot-gun.

rightsaidjames:
@carlmw I agree with the decision to make the button/link a non-interactive element when disabled, but I'm wondering if a `div` is the correct element to choose here (instead of `span`) given that it's a container element which holds divs and images?

carlmw:
> @carlmw I agree with the decision to make the button/link a non-interactive element when disabled, but I'm wondering if a `div` is the correct element to choose here (instead of `span`) given that it's a container element which holds divs and images?

you're entirely correct. I have this archaic notion in my head that `<a>` is an `"inline"` element which means that it should naturally be replaced with another inline element so I automatically reach for a `<span>`. Indeed it used to be the case that you couldn't put a block element inside an inline but they changed `<a>` with HTML5 IIRC. Thanks for spotting this 🙇🏼

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
kimon-satan: Yeah  I can imagine that this component is going to get quite silly at some point as we introduce more sections to lessons. At that point we can just refactor out a generalised nav component as you suggest below and then move this component over to OWA

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
kimon-satan: That could be useful for the future . (see my previous comment)

src/components/integrated/OakLessonNavItem/OakLessonNavItem.tsx
kimon-satan: I will try to explain to you on a phone call and see if I'm talking rubbish.

kimon-satan (approved):

PR #73, opened 2024-01-22
fix(PUPIL-78): replace deprecated names for row and column gap

carlmw:
`grid-column-gap` and `grid-row-gap` are deprecated names for `column-gap` and `row-gap`. They're still respected as aliases for the new property names but I suspect that is why they didn't appear in your test.


kimon-satan (approved):
awesome thanks

PR #75, opened 2024-01-22
Feat/add form id to button

benprotheroe:
add form prop to internalButton component so that submit buttons can live outside of the form they affect

### DIFF
--- a/src/components/base/InternalButton/InternalButton.tsx
+++ b/src/components/base/InternalButton/InternalButton.tsx
@@ -90,7 +90,7 @@ const UnstyledInternalButton = (props: InternalButtonProps) => {
       onMouseEnter={handleMouseEnter}
       onMouseLeave={handleMouseLeave}
       type={props.type ?? "button"}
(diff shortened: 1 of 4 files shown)


src/components/base/InternalButton/InternalButton.tsx
```diff
@@ -89,6 +90,7 @@ const UnstyledInternalButton = (props: InternalButtonProps) => {
       onMouseEnter={handleMouseEnter}
       onMouseLeave={handleMouseLeave}
       type={props.type ?? "button"}
+      form={props.form ? props.form : undefined}
```
carlmw, line 93:
```suggestion
      form={props.form}
```

ternary seems redundant

carlmw (approved):
lgtm, don't think you need that ternary though, the prop can be forwarded as is.

src/components/base/InternalButton/InternalButton.tsx
benprotheroe: oh yeah

PR #76, opened 2024-01-22
Fix/pupil 312/mcq feedback states

kimon-satan:
### Acceptance criteria

- [x]  Correctly selected answer is green and checked
- [x]  Incorrectly selected answer is red and crossed
- [x]  Incorrectly unselected answer is white and checked
- [x]  correctly unselected is a neutral state
- [x]  Stories should be updated with correct labels to prevent any future confusion.

https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=3473%3A12270&mode=dev

### To test
https://deploy-preview-76--lively-meringue-8ebd43.netlify.app/?path=/story/components-integrated-oakquizradiobutton--with-feedback
https://deploy-preview-76--lively-meringue-8ebd43.netlify.app/story/components-integrated-oakquizcheckbox--feedback

### DIFF
--- a/src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
+++ b/src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
@@ -157,14 +157,20 @@ export const OakQuizCheckBox = (props: OakQuizCheckBoxProps) => {
     disabled,
     innerRef,
     displayValue,
+    defaultChecked,
     ...rest
   } = props;
 
   const isFeedback = !!feedback;
-  const isCorrect = feedback === "correct";
 
   const defaultRef = useRef<HTMLInputElement>(null);
   const inputRef = innerRef ?? defaultRef;
+  const showTick =
+    (feedback === "correct" && (inputRef.current?.checked || defaultChecked)) ||
+    (feedback === "incorrect" && !inputRef.current?.checked && !defaultChecked);
+  const showCross =
+    feedback === "incorrect" && (inputRef.current?.checked || defaultChecked);
+
   const handleContainerClick = (
     e:
       | React.MouseEvent<HTMLDivElement>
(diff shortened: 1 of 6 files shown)


src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
```diff
@@ -157,14 +157,20 @@ export const OakQuizCheckBox = (props: OakQuizCheckBoxProps) => {
...
-  const isCorrect = feedback === "correct";
 
   const defaultRef = useRef<HTMLInputElement>(null);
   const inputRef = innerRef ?? defaultRef;
+  const showTick =
+    (feedback === "correct" && (inputRef.current?.checked || defaultChecked)) ||
+    (feedback === "incorrect" && !inputRef.current?.checked && !defaultChecked);
+  const showCross =
+    feedback === "incorrect" && (inputRef.current?.checked || defaultChecked);
+
```
carlmw, line 173: My gut is telling me there is a bug in here. Do you need to reference `defaultChecked`, is it enough to read the `checked` state from the ref?

I reckon this component could get in a state where the feedback is out of sync with the checkbox. `defaultChecked` is the initialising state for the checkbox but you're always referencing it here so in some cases it will trump the current `checked` state of the checkbox. 

If you can remove `defaultChecked` altogether I would. Failing that i'd make the internal checkbox controlled and manage its checked state with `useState` .

Something like:

```typescript
const [isChecked, setIsChecked] = useState(defaultChecked);
const showTick =
    (feedback === "correct" && isChecked) ||
    (feedback === "incorrect" && !isChecked);
  const showCross =
    feedback === "incorrect" && isChecked);
const handleChange: ChangeEventHandler<HTMLInputElement> = (event) => {
  setIsChecked(event.currentTarget.checked);
  if (onChange) onChange(event);
}
```

More code, but it sees off a subtle bug that would be hard to identify down the road. Hope that makes sense.

carlmw (commented):
Not sure I can make sense of the requirements anymore 😢 but there is something about that `defaultChecked` prop that is subtly problematic.

src/components/integrated/OakQuizCheckBox/OakQuizCheckBox.tsx
kimon-satan: It makes sense but changes so much of the implementation higher up that I really don't want to do it at this stage. DefaultChecked is basically only used for the StoryBook so I'm wondering whether the solution is just to not expose and use defaultChecked at all.

kimon-satan: Which is of course what you said 🙄

carlmw (approved):
Thanks, I had a feeling it was only for Storybook.

Going forward we should probably always add a `checked` prop as well as `defaultChecked` so that it matches the API of the underlying `<input>` allowing it to be used controlled and uncontrolled.

PR #80, opened 2024-01-22
feat(PUPIL-24): Serve images directly from Cloudinary

carlmw:


.env.example
```diff
@@ -1,2 +1,3 @@
 NEXT_PUBLIC_OAK_ASSETS_HOST=
 NEXT_PUBLIC_OAK_ASSETS_PATH=
+NEXT_PUBLIC_CLOUDINARY_CLOUD_NAME=
```
carlmw, line 3: Annoyingly this is needed by `next-cloudinary`. It's the first part of `NEXT_PUBLIC_OAK_ASSETS_PATH` so perhaps there is an opportunity to remove `NEXT_PUBLIC_OAK_ASSETS_PATH` since the `/image/upload` portion appears to be constant to all Cloudinary accounts?

I guess this will be a breaking change

src/components/base/OakImage/OakImage.test.tsx
```diff
@@ -47,4 +47,19 @@ describe("OakImage", () => {
...
+    it("uses the Cloudinary Image component", () => {
+      const { getByRole } = render(
+        <OakImage
+          src="https://res.cloudinary.com/mock-cloudinary-cloud/image/upload/v1608635688/image.jpg"
+          alt="a test image"
+        />,
+      );
+
+      expect(getByRole("img").getAttribute("src")).toEqual(
+        "https://res.cloudinary.com/mock-cloudinary-cloud/image/upload/c_limit,w_3840/f_auto/q_auto/v1608635688/image?_a=BAVAfVAO0",
```
carlmw, line 61: These are the transformation URLs that `CldImage` generates from the original URL we feed in.

src/components/integrated/OakHintButton/__snapshots__/OakHintButton.test.tsx.snap
```diff
@@ -188,7 +188,7 @@ exports[`OakHintButton matches snapshot 1`] = `
             loading="lazy"
             onError={[Function]}
             onLoad={[Function]}
-            src="https://undefined/undefined/v1705078631/icons/zldisxmbff36z68rwcef.svg"
+            src="https://res.cloudinary.com/mock-cloudinary-cloud/image/upload/v1705078631/icons/zldisxmbff36z68rwcef.svg"
```
carlmw, line 191: Nice to be rid of those `undefined/undefined` in the snapshots

src/test-helpers/setupTests.ts
```diff
@@ -1 +1,5 @@
 import "jest-styled-components";
+
+process.env.NEXT_PUBLIC_OAK_ASSETS_HOST = "res.cloudinary.com";
+process.env.NEXT_PUBLIC_OAK_ASSETS_PATH = "mock-cloudinary-cloud/image/upload";
+process.env.NEXT_PUBLIC_CLOUDINARY_CLOUD_NAME = "mock-cloudinary-cloud";
```
carlmw, line 5: I was initially going to add a `.env.test` but tbh that is more stuff a dev has to deal with and we'd have to mess with CI too.

kimon-satan (commented):
I've added the env var to Netlify but the error still shows on the preview. Could it be that the env var needs to be added to .storybook/main.ts as well ?
Some other thoughts:
- Cloudinary is doing something rotten to SVGs. We can avoid serving them through cloudinary when we know that they are svgs eg icons (I'd like to see that on the storybook for reassurance) but we also need to check whether quizzes have svgs
- In OWA we have two cloudinary accounts: One for quiz images and one for icons and app assets. How can we deal with this ?

carlmw:
@kimon-satan 

All up and running now, you were correct it needed adding to `main.ts`

1. I've added a test that will skip the optimisations if the file ends in `.svg` — so it will serve SVGs up directly, this includes icons from the assets account and any quiz images that happen to be SVGs. The `f_auto` transform (which we could remove) is Cloudinary's way of trying to be smart and picking the best format for the asset — in most cases it will serve up a webp unless the browser is old. I would have thought leaving SVGs as is would make sense but there is probably some business reason they do this.
2. It probably makes sense to only pass quiz images through this optimisation process since they're most likely to huge and we can be sure that app assets and icons are already well optimised. We can tweak the config to point at the quiz assets Cloudinary account and only those images will receive the treatment.

How does that sound?

src/components/base/OakCloudinaryImage/OakCloudinaryImage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  cloudinaryId: string;
+};
+
+/**
+ * OakCloudinaryImage wraps OakImage providing responsive images from Cloudinary
+ * based on the `sizes` prop.
+ *
+ * Cloudinary config can be set globally with `setCloudinaryConfig` or per image with the `config` prop.
+ */
+export const OakCloudinaryImage = ({
```
carlmw, line 30: I figured a change in tack was needed. The Cloudinary processing should be opt-in since most of our images are UI elements and icons which should already be optimised/served locally.

Adding this wrapper component makes it explicit that the image is expected to be served from Cloudinary with all the bells and whistles of responsive sizing.

src/components/base/OakCloudinaryImage/OakCloudinaryImage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  config: incomingConfig,
+  cloudinaryId,
+  unoptimized = false || cloudinaryId.endsWith(".svg"),
+  ...props
+}: OakCloudinaryImageProps) => {
+  const config = incomingConfig ?? getCloudinaryConfig();
```
carlmw, line 36: This allows us to serve images from multiple Cloudinary instances (in our case the assets or quiz images).

a global/default config can be set with `setCloudinaryConfig` or a prop passed with a specific config for a given image.

carlmw:
@kimon-satan i've changed my approach which makes the change a little less disruptive to `OakImage` — i've gone ahead and created a special `OakCloudinaryImage` which can be used to opt-in to Cloudinary image processing.

This feels better to me and supports the multiple Cloudinary instance use-case that we have.

src/components/base/OakCloudinaryImage/cloudinary.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function getCloudinaryConfig() {
+  if (!cloudinaryConfig) {
+    throw new Error(
+      "Cloudinary config is not set for `OakCloudinaryImage`, use `setCloudinaryConfig` to set it.",
+    );
+  }
+  return cloudinaryConfig;
+}
```
carlmw, line 16: 🚿 thoughts from last night. Coming back to this today and this should definitely be a provider. I'll make it so.

PR #86, opened 2024-01-26
PUPIL-325: Add `OakHandDrawnCard` and supporting internal components

carlmw:
Introduces `OakHandDrawnCard` which will initially be used to add the call out text at the bottom of the [lesson review page](https://www.figma.com/file/G7z5ufo2T0CHk14CkadeuB/Pupils-high-fidelity-designs?type=design&node-id=3952-185829&mode=dev), but will also drive the [subject icon on the lesson intro page](https://www.figma.com/file/G7z5ufo2T0CHk14CkadeuB/Pupils-high-fidelity-designs?type=design&node-id=3072-14821&mode=dev) (plus more hand drawn looking elements to come)

https://deploy-preview-86--lively-meringue-8ebd43.netlify.app/?path=/docs/components-ui-oakhanddrawncard--docs
https://deploy-preview-86--lively-meringue-8ebd43.netlify.app/?path=/docs/components-base-internalcardwithbackgroundelement--docs
https://deploy-preview-86--lively-meringue-8ebd43.netlify.app/?path=/docs/components-base-internalstyledsvg--docs

Screenshot 2024-01-26 at 14 32 06
Screenshot 2024-01-26 at 12 50 49
Screenshot 2024-01-26 at 12 50 37


src/components/ui/OakHandDrawnCard/OakHandDrawnCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+            vectorEffect="non-scaling-stroke"
```
carlmw, line 49: This is a really useful SVG feature, it means that the `stroke-width` will not scale with the rest of the image. Instead it will be calculated relative to the page — so a 4px stroke will be a 4px stroke on the page even if the SVG has been scaled up or down. Really useful in this case when we want to stretch a shape to fill an arbitrary space while applying a border that matches a design token.

src/styles/helpers/parseBorderWidth.ts
```diff
@@ -0,0 +1,15 @@
...
+  value: OakBorderWidthToken | null | undefined,
+) {
+  if (!value) {
+    return undefined;
+  }
+
+  if (value in oakBorderWidthTokens) {
+    return `${pxToRem(oakBorderWidthTokens[value])}rem`;
+  }
+}
```
carlmw, line 15: `parseBorder` also applies a `border-style` E.g `0.125rem solid` which won't work with `stroke-width`. So i've added `parseBorderWidth` which follows the same conventions but only returns the size.

src/styles/utils/positionStyle.ts
```diff
@@ -23,6 +23,7 @@ export type PositionStyleProps = {
   $right?: ResponsiveValues<PositionSpacing>;
   $bottom?: ResponsiveValues<PositionSpacing>;
   $left?: ResponsiveValues<PositionSpacing>;
+  $inset?: ResponsiveValues<PositionSpacing>;
```
carlmw, line 26: No positioning toolkit is complete without `inset`

benprotheroe: i think my prs need this level of description

src/components/base/InternalStyledSvg/InternalStyledSvg.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * A styled SVG element with props to apply design tokens to the fill and stroke.
+ */
+export const InternalStyledSvg = styled.svg<InternalStyledSvgProps>`
+  ${responsiveStyle<InternalStyledSvgProps, OakCombinedColorToken>(
+    "fill",
+    (props) => props.$fill,
+    parseColor,
+  )}
```
benprotheroe, line 25: i feel liket his slightly breaks our convention that "styles" live in the style/utils folder and then they are imported into the base components; rather then putting the style logic inside the base components file

benprotheroe (approved):

PR #88, opened 2024-01-29
PUPIL-344 make button related components polymorphic

weronika-szalas:
Convert button related components into polymorphic components so they can use any tag as a base tag. This will allow us to use them as links styled like buttons without duplicating the code.

Added relevant stories to Storybook.

### DIFF
--- a/src/components/base/InternalButton/InternalButton.test.tsx
+++ b/src/components/base/InternalButton/InternalButton.test.tsx
@@ -82,7 +82,7 @@ describe("InternalButton", () => {
   });
 
   it("correctly fires for a form matching the id from its form props", () => {
-    const onSubmit = jest.fn();
+    const onSubmit = jest.fn((e) => e.preventDefault());
     const { getByRole } = render(
       <div>
         <form id="test-form" onSubmit={onSubmit}>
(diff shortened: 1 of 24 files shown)


rightsaidjames (approved):
Looks good - I've done some manual accessibility testing on all of the new/updated components and all seems hunky-dory!

My only concern, which in practice is unlikely to be an issue given the accessibility knowledge of our devs, is whether allowing a button to be _any_ element, rather than a preset list of elements, gives too much control and could lead to elements that are intended to be interactive but lack the necessary semantics and keyboard controls. But I've recently spoken to @carlmw about the (valid) use case of outputting a disabled link as a div (to ensure that it has no link semantics and is not interactive), so I don't see why the same couldn't apply to buttons. Additionally, I'm pretty certain any misuse of this component when implemented into OWA would be spotted at PR review anyway!

src/components/base/InternalButton/InternalButton.test.tsx
```diff
@@ -82,7 +82,7 @@ describe("InternalButton", () => {
   });
 
   it("correctly fires for a form matching the id from its form props", () => {
-    const onSubmit = jest.fn();
+    const onSubmit = jest.fn((e) => e.preventDefault());
```
carlmw, line 85: thanks

carlmw (approved):
This is a very good precedent to set.

PR #91, opened 2024-01-31
PUPIL-283: Tweaks for lesson overview

carlmw:
Just a couple of minor tweaks to the lesson layout and nav item components.

### TODO after publishing
- [ ] Review existing pages that use `OakLessonLayout` and re-apply the padding in the consuming component.

### DIFF
--- a/src/components/integrated/OakLessonLayout/OakLessonLayout.tsx
+++ b/src/components/integrated/OakLessonLayout/OakLessonLayout.tsx
@@ -119,9 +119,9 @@ function pickSectionColours(
   switch (sectionName) {
     case "overview":
       return [
-        "bg-decorative4-subdued",
+        "bg-decorative1-main",
         "bg-primary",
-        "border-decorative4",
+        "border-decorative1-stronger",
         "bg-primary",
       ];
     case "intro":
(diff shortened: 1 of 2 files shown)


src/components/integrated/OakLessonLayout/OakLessonLayout.tsx
```diff
@@ -86,11 +86,7 @@ export const OakLessonLayout = ({
             {topNavSlot}
           </OakBox>
         )}
-        <OakFlex
-          $flexGrow={1}
-          $flexDirection="column"
-          $mh={["space-between-s", "space-between-l"]}
```
carlmw, line 92: This was interfering with the layout for the lesson overview — at the mobile breakpoint the subject icon and lesson name become a header to the content. The bottom border runs to the edge of the viewport which isn't possible with the original implementation (unless I employed negative margins 😉 )

Screenshot 2024-01-31 at 17 46 44

This also interfered with the grid since the columns would be inset my the padding.

Easier to remove the automatic padding and leave it up to the consumer to apply a proper grid.

I'll go back and correct things where the layout has already been used and is reliant on the padding

src/components/integrated/OakLessonLayout/OakLessonLayout.tsx
```diff
@@ -123,9 +119,9 @@ function pickSectionColours(
   switch (sectionName) {
     case "overview":
       return [
-        "bg-decorative4-subdued",
+        "bg-decorative1-main",
         "bg-primary",
-        "border-decorative4",
+        "border-decorative1-stronger",
         "bg-primary",
```
carlmw, line 125: The overview page is no longer violet, but green — this is to soften the transition from the old design.

benprotheroe (approved):

PR #92, opened 2024-02-01
feat: create OakPrimaryNavItem and OakPrimaryNav components

weronika-szalas:
Create `OakPrimaryNavItem` and `OakPrimaryNav` components

`OakPrimaryNav`:
https://deploy-preview-92--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oakprimarynav--docs

`OakPrimaryNavItem`:
https://deploy-preview-92--lively-meringue-8ebd43.netlify.app/?path=/docs/components-integrated-oakprimarynavitem--docs

Mobile view of this navigation will be a separate component and it will be controlled with `display: none` or `display: block` to display relevant version depending on the viewport.

Screenshot 2024-02-01 at 14 40 30

### DIFF
--- a/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
+++ b/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
@@ -24,14 +24,14 @@ export const OakPrimaryNav = ({ ariaLabel, navItems }: OakPrimaryNavProps) => {
         $ph={"inner-padding-none"}
       >
         {navItems.map(
-          ({ href, isCurrent, label, ...rest }: OakPrimaryNavItemProps) => (
-            <OakLI $listStyle={"none"} key={label}>
-              <OakPrimaryNavItem
-                href={href}
-                isCurrent={isCurrent}
-                label={label}
-                {...rest}
-              />
+          (
+            { href, isCurrent, children }: OakPrimaryNavItemProps,
+            index: number,
+          ) => (
+            <OakLI $listStyle={"none"} key={index}>
+              <OakPrimaryNavItem href={href} isCurrent={isCurrent}>
+                {children}
+              </OakPrimaryNavItem>
             </OakLI>
           ),
         )}
--- a/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
+++ b/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
@@ -15,15 +15,17 @@ export type OakPrimaryNavProps = {
 export const OakPrimaryNav = ({ ariaLabel, navItems }: OakPrimaryNavProps) => {
   return (
     <nav aria-label={ariaLabel}>
-      <OakFlex as={"ul"}>
+      <OakFlex
+        as={"ul"}
+        $flexDirection={["row"]}
+        $gap={"space-between-m"}
+        $mh={"space-between-none"}
+        $mv={"space-between-none"}
+        $ph={"inner-padding-none"}
+      >
         {navItems.map(
           ({ href, isCurrent, label, ...rest }: OakPrimaryNavItemProps) => (
-            <OakLI
-              $listStyle={"none"}
-              $mr={["space-between-m", "space-between-m", "space-between-none"]}
-              $mb={["space-between-m", "space-between-m", "space-between-none"]}
-              key={label}
-            >
+            <OakLI $listStyle={"none"} key={label}>
               <OakPrimaryNavItem
                 href={href}
                 isCurrent={isCurrent}
--- a/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
+++ b/src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
@@ -1,0 +1,39 @@
+import React from "react";
+
+import { OakFlex } from "@/components/base/OakFlex";
+import { OakLI } from "@/components/base";
+import {
+  OakPrimaryNavItem,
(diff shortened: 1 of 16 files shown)


src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const OakPrimaryNav = ({ ariaLabel, navItems }: OakPrimaryNavProps) => {
+  return (
+    <nav aria-label={ariaLabel}>
+      <OakFlex as={"ul"}>
+        {navItems.map(
+          ({ href, isCurrent, label, ...rest }: OakPrimaryNavItemProps) => (
+            <OakLI
+              $listStyle={"none"}
+              $mr={["space-between-m", "space-between-m", "space-between-none"]}
```
kimon-satan, line 23: this doesn't seem to make sense . We end up with larger margins on the side and the bottom for the mobile view whilst none for desktop ? Also we end up with some unnecessary margin on the rhs after the final element this way. Could we instead use the $gap property of parent flex and also the flex direction ?

kimon-satan (changes requested):
Just one comment about the margins

src/components/integrated/OakPrimaryNav/OakPrimaryNav.tsx
weronika-szalas: updated

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {label}
+    </OakPrimaryButton>
+  ) : (
+    <OakPrimaryInvertedButton
+      element={Link}
+      href={href}
+      shallow={shallow}
+      isTrailingIcon={true}
+      aria-label={label}
+      aria-disabled={isCurrent}
```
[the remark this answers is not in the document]
weronika-szalas: yes, I wanted to discuss this with you - it's either above or we can make it a span element

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    >
+      {label}
+    </OakPrimaryButton>
+  ) : (
+    <OakPrimaryInvertedButton
+      element={Link}
+      href={href}
+      shallow={shallow}
+      isTrailingIcon={true}
+      aria-label={label}
```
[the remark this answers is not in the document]
weronika-szalas: updated

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    >
+      {label}
+    </OakPrimaryButtonAriaDisabled>
+  ) : (
+    <OakPrimaryInvertedButton
+      element={Link}
+      href={href}
+      shallow={shallow}
+      isTrailingIcon={true}
+      aria-disabled={isCurrent}
```
rightsaidjames, line 58: Is this attribute no longer needed now that we're handling the disabled state without ARIA? Its value can now only be false, as far as I can tell.

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const OakPrimaryNavItem = ({
+  href,
+  isCurrent = false,
+  shallow = true,
+  label,
+  ...rest
+}: OakPrimaryNavItemProps) => {
+  return isCurrent ? (
+    <OakPrimaryButtonAriaDisabled
```
rightsaidjames, line 45: `Aria` is no longer being used to make the link inactive, so does the component need renaming?

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
weronika-szalas: removed

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
weronika-szalas: renamed

rightsaidjames (approved):
Lookin' good

src/components/integrated/OakPrimaryNavItem/OakPrimaryNavItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+  OakPrimaryInvertedButtonProps,
+} from "@/components/ui/OakPrimaryInvertedButton";
+
+type Url = string | UrlObject;
+
+export type OakPrimaryNavItemProps = {
+  href: Url;
+  isCurrent?: boolean;
+  shallow?: boolean;
+  label: string;
```
carlmw, line 22: minor, but this should probably be `children` and the type `ReactNode` to fit into the components it's wrapping. As a consumer i'd expect the API to be the same as a regular button/link.

carlmw (approved):
No blockers for me just a minor. As discussed in our call I reckon this component can be decomposed a little later for re-use elsewhere 👍🏼 :shipit: .

kimon-satan (approved):

PR #97, opened 2024-02-05
Fix/pupil 340/quiz text input widths

kimon-satan:
### Story
The quiz text input should remain the same size when switching between input and feedback states . Currently it expands to accommodate the icon.

comments from designer:

“For the 2nd question of the starter quiz, and the 1st question of the exit quiz there is some shift in the size of the text input when submitting the question on desktop, and when the icon appears when in a valid or invalid state. The text input width shouldn't increase in width ideally across states. I think the width should be increased to something like 500px as a default for desktop. The text input should span the full width for tablet and mobile.” 

### Implementation details

- Modify the underlying textInput so that setting the width adjusts the wrapper width rather than attempt (and fail) to adjust the input width
- change width prop on this component to wrapperWidth so that the intention is clear but also there are no type clashes

### Testing 
https://deploy-preview-97--lively-meringue-8ebd43.netlify.app/?path=/story/components-integrated-oakquiztextinput--responsive-width
https://deploy-preview-97--lively-meringue-8ebd43.netlify.app/?path=/docs/components-ui-oaktextinput--docs

### Acceptance criteria

- [x]  **OakTextInput** should be able to set to 100% of the parent width
- [x] **OakTextInput**  & **OakQuizTextInput**  should accept responsive values for wrapperWidth

### DIFF
--- a/src/components/integrated/OakQuizTextInput/OakQuizTextInput.stories.tsx
+++ b/src/components/integrated/OakQuizTextInput/OakQuizTextInput.stories.tsx
@@ -40,3 +40,13 @@ export const WithIncorrectFeedback: Story = {
     <OakQuizTextInput value="An incorrect answer" feedback="incorrect" />
   ),
 };
+
+export const ResponsiveWidth: Story = {
+  render: () => (
+    <OakQuizTextInput
+      value="An incorrect answer"
+      feedback="incorrect"
+      wrapperWidth={["100%", "all-spacing-22"]}
+    />
+  ),
+};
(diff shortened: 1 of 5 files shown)


Monika-Koziol:
@kimon-satan how can I preview it?

Monika-Koziol (approved):
The size of the input looks fine now. I know this is not part of this ticket but should we show truncation to the user instead of cutting off text? Or even allow for the text box to grow if the answer is long.

carlmw (approved):

kimon-satan:
> The size of the input looks fine now. I know this is not part of this ticket but should we show truncation to the user instead of cutting off text? Or even allow for the text box to grow if the answer is long.

will make a ticket

PR #99, opened 2024-02-05
fix: focus ring when nested in list

kimon-satan:
### User Story

This happens with OakPrimary and secondary buttons when browsing on safari

image

Happens when using buttons within OakLI and OakUL

Fix:

cool bug, sticking `top:0` on both of the boxes that apply the focus ring will fix it. Seems to have something to do with the immediate descendent of the component being `display: list-item` , i'm guessing it has some impact on how the `inset` is inherited.

### Acceptance criteria

- [x]  Buttons can be use in OakUL / OakLI

### Testing

http://localhost:6006/?path=/story/components-ui-internalshadowrectbutton--button-in-list
http://localhost:6006/?path=/story/components-ui-internalshadowroundbutton--button-in-list

Have a look at any of the other buttons too. They're all effected by these changes


benprotheroe (approved):
can see any examples of this but will take your work for it

PR #103, opened 2024-02-12
feat: add `grid-template-rows` prop to OakGrid component

carlmw:
this is a missing piece that lets us do a lot more with grid layouts.

I need this to properly layout the copyright notice on the lesson intro page.


kimon-satan (approved):
nice

PR #108, opened 2024-02-16
feat(PUPIL-274): add missing sizing keywords and set a default button width

carlmw:
* `min-content` and `max-content` are very useful.
* `max-content` seems like a sensible default for button widths (to prevent text wrapping)
  * we should be able to safely apply this since the buttons are only currently used in the pupil section and none have intentional wrapping
* I will use this to prevent wrapping in `OakLessonBottomNav`

Screenshot 2024-02-16 at 12 24 43

Screenshot 2024-02-16 at 12 26 47

Screenshot 2024-02-16 at 12 30 12


kimon-satan (approved):
sounds like a good choice to me

PR #113, opened 2024-02-22
chore(PUPIL-439): add some basic Oak branding to Storybook

carlmw:
https://deploy-preview-113--lively-meringue-8ebd43.netlify.app/

Before we make this public and to lend a little more polish to the documentation we're adding some basic branding to Storybook. We can't do much more than this without getting under the skin of Storybook.

* I had a look at the brand guidelines and I think this is in keeping with them
* Applies oak colours and Lexend font within the constraints of Storybook's basic theming options.

Screenshot 2024-02-22 at 16 44 25

Using a mint version of the favicon so that it is easier to distinguish between tabs
Screenshot 2024-02-22 at 16 44 50


messmike:
Looking good, thanks @carlmw! Adding @Monika-Koziol so you can take a look.

messmike (approved):

messmike:
hey @carlmw sorry I know this is now merged. We have one minor styling thing for this, but it's not a major one. If possible could we use a lighter grey for the borders `(border-neutral-lighter)`, or even set them to 'none' so they end up looking more like this:

messmike:
Screenshot 2024-02-26 at 3 06 36 pm

PR #137, opened 2024-03-27
ENG-692 Add `OakModal`

carlmw:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

https://github.com/oaknational/oak-components/assets/122096/a3f6a60c-109c-41f6-9824-5c1cdddd118f

## Link to the design doc

https://www.figma.com/file/8uwqYUIpTpyqdpSAVVV0g9/Oak-DS-Components?type=design&node-id=847-918612&mode=dev

https://deploy-preview-137--lively-meringue-8ebd43.netlify.app/?path=/docs/components-molecules-oakmodal--docs

## ACs
- [x]  Takes a11y cues from https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/
    - [x]  Keyboard support
    - [x]  Focus locked within the modal when open
    - [x]  Relevant aria roles applied
- [x]  Responsive at the 3 Oak breakpoints, mobile, tablet and desktop
- [x]  Exported from oak-components as a molecule
- [x]  Animates in from the left?
- [x]  Footer sticks to the bottom of the viewport
    - [x]  Body can scroll
    - [x]  “Background” page cannot be scrolled while modal is active


package.json
```diff
@@ -80,7 +82,9 @@
     "prettier": "^3.1.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-focus-on": "^3.9.2",
```
carlmw, line 85: This is a one-stop-shop for applying good a11y to modal UI

https://github.com/theKashey/react-focus-on

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+      nodeRef={transitionRef}
+      addEndListener={(done) => {
+        transitionRef.current?.addEventListener("transitionend", done);
+      }}
+      timeout={500}
+      mountOnEnter
+      unmountOnExit
+    >
+      {(state) => (
+        <FocusOn onEscapeKey={onClose} returnFocus autoFocus>
```
carlmw, line 122: `returnFocus` will restore focus to the element that was active before the modal was opened (most likely the button that opened the modal in the first place) this ensures that the user does not lose their place in the document

`autoFocus` will give focus to a focusable element inside the modal — typically the first element in the tab index.

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+                disabledTextColor="transparent"
+                hoverIconBackground="black"
+                hoverIconColor="white"
+                disabledIconBackground="transparent"
+                iconBackgroundSize="all-spacing-6"
+                iconSize="all-spacing-6"
+                iconName="cross"
+              />
+            </OakFlex>
+            <div style={{ display: "contents" }} data-autofocus-inside>
```
carlmw, line 175: `data-autofocus-inside` is a hook that `react-focus-on` will detect and scope it's initial `autoFocus` target to the first element in the tab index that is inside this element.

src/components/molecules/InternalShadowRoundButton/InternalShadowRoundButton.tsx
```diff
@@ -151,6 +159,7 @@ export const InternalShadowRoundButton = <C extends ElementType = "button">(
                 ? defaultIconColor
                 : null
           }
+          data-icon-for="button"
```
carlmw, line 162: A styling hook to target the icon from a parent styled-component. It's needed since there is no non-awful way to target hover or focus styles from a parent context without a bunch of JavaScript. It breaks the encapsulation of styles but perhaps CSS custom properties would be a better approach (though we haven't used them anywhere yet!)

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+                defaultTextColor="transparent"
+                hoverTextColor="transparent"
+                disabledTextColor="transparent"
+                hoverIconBackground="black"
+                hoverIconColor="white"
+                disabledIconBackground="transparent"
+                iconBackgroundSize="all-spacing-6"
+                iconSize="all-spacing-6"
+                iconName="cross"
+              />
```
carlmw, line 173: This button has a new component smell about it — but I wouldn't want to introduce a new flavour of button until there is another example of it in the wild. This might just be a one off "cross" to close the modal after all and the `Oak*` namespace is getting muddy already.

package.json
```diff
@@ -80,7 +82,9 @@
...
     "react-test-renderer": "^18.2.0",
+    "react-transition-group": "^4.4.5",
```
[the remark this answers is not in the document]
kimon-satan: How big is it ? Might make life easier to keep it independent of OWA

src/components/molecules/InternalShadowRoundButton/InternalShadowRoundButton.tsx
kimon-satan: I don't mind this approach as it's clear what these are for. I can see things getting awfully muddled if we introduce css variables as people might start using them instead of the existing token system.

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+ */
+export const OakModal = ({
+  children,
+  footerSlot,
+  domContainer = document.body,
+  isOpen,
+  onClose,
+  ...rest
+}: OakModalProps) => {
+  const [canaryElement, setCanaryElement] = useState<HTMLDivElement | null>(
```
kimon-satan, line 85: So the canaryElement is set to an empty div which can be used to attach the IntersectionObserver. This is inturn is used to update isScrolled state  ?

kimon-satan (approved):
No blockers from me .

src/components/molecules/OakModal/OakModal.tsx
carlmw: that is correct. The canary sits at the top of the element and when it scrolls out of/into view (I.e it no longer / now intersects the viewport) the observer fires and we can use that to cheaply detect that the element has been scrolled.

package.json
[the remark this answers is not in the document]
carlmw: @kimon-satan about 6KB of the bundle gzipped. Not huge. I guess ideally OWA wouldn't have need for react-transition-group after full adoption of oak-components. 

Screenshot 2024-03-27 at 17 19 06

src/components/molecules/InternalShadowRoundButton/InternalShadowRoundButton.tsx
carlmw: 🤔 possibly though the advantage of CSS custom props would be that you could open up specific properties to target from an ancestor which would carve a tightly controlled API across the component boundary.

E.g with this example there is a very clear contract to be maintained through future changes.

```html
<button>
  <img src="foo.png" data-icon-for="button" />
</button>
```

```css
img {
  filter: var(--oak-icon-color-filter, ${(props) => parseColorFilter(props.$colorFilter)});
}

button:hover {
  --oak-icon-color-filter: ${parseColorFilter('black')};
}
```

The current approach leaves specificity up to chance and you can target any property (or add arbitrary properties!)

```css
img {
  filter: ${(props) => parseColorFilter(props.$colorFilter)};
}

button:hover [data-icon-for="button"] img {
  filter: ${parseColorFilter('black')};
  background-color: orange!important;
}
```

package.json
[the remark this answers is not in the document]
kimon-satan: Cool . Looks reasonable enough to keep it in then.

PR #169, opened 2024-05-13
Chore/upgrade storybook [LESQ-821]

njblore:
Upgrade to v8 of storybook and its dependencies. Required updating nextjs to v13.5 as well.
Mostly just small type changes, in particular the controlType in `argTypes` for stories:
- type `select` is the default and can be omitted
-`type` key is not required for most control types
- except for types with additional config (eg 'number', 'range') the `type` key is required and I had to cast the type as const here, the actual type (for `type`...) doesn't seem to be exported from the library

https://deploy-preview-169--lively-meringue-8ebd43.netlify.app/
  
## Testing instructions
Check stories are still building and displaying as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This is a React Typescript components library which supports React and Next appl
 
 ## Installation
 
-This library is suitable for use in an app using React 18 and Next.js 13+
+This library is suitable for use in an app using React 18 and Next.js 13.5+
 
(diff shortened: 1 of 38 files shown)


kimon-satan (approved):
Oh god that's a bit of annoying one. NB. when you are merge there might be more stories with gotchyas in as we've been doing work on OC this morning

PR #202, opened 2024-06-18
fix: `OakModal` blowing up when doing SSR

carlmw:
`document.body` isn't available during SSR, so this would blow up.

This commit adds an awful `useEffect` which flips an `isMounted` boolean which we can use to determine if we're rendering on the client while avoiding a server->client markup mismatch warning. *groan*

# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

https://deploy-preview-202--lively-meringue-8ebd43.netlify.app/?path=/docs/components-molecules-oakmodal--docs

## Testing instructions

This should behave as before in Storybook where the modal is always rendered client-side. The only way to test this fix is to use it with SSR (which i've done in OWA to verify).

### DIFF
--- a/src/components/molecules/OakModal/OakModal.test.tsx
+++ b/src/components/molecules/OakModal/OakModal.test.tsx
@@ -20,16 +20,26 @@ jest.mock("react-dom", () => {
 });
 
 describe(OakModal, () => {
-  it("matches snapshot", () => {
+  it("does not render until mounted on the client", () => {
     const tree = create(
       <OakThemeProvider theme={oakDefaultTheme}>
         <OakModal isOpen onClose={() => {}} footerSlot="Modal footer">
           Modal content
         </OakModal>
       </OakThemeProvider>,
-    ).toJSON();
+    );
 
-    expect(tree).toMatchSnapshot();
+    expect(tree.toJSON()).toBeNull();
+  });
+
+  it("matches snapshot when mounted", async () => {
+    const result = renderWithTheme(
+      <OakModal isOpen onClose={() => {}}>
+        Modal content
+      </OakModal>,
+    );
+
+    expect(result.container).toMatchSnapshot();
   });
 
   it("calls onClose when the close button is clicked", () => {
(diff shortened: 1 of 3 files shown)


src/components/molecules/OakModal/OakModal.test.tsx
```diff
@@ -20,16 +20,26 @@ jest.mock("react-dom", () => {
...
+
+  it("matches snapshot when mounted", async () => {
+    const result = renderWithTheme(
+      <OakModal isOpen onClose={() => {}}>
+        Modal content
+      </OakModal>,
+    );
 
-    expect(tree).toMatchSnapshot();
+    expect(result.container).toMatchSnapshot();
```
carlmw, line 42: Waiting for the shallow renderer to run hooks (which is necessary now that the component is not rendered immediately) introduced this error https://github.com/theKashey/react-focus-lock/issues/123 from `react-focus-lock` which I couldn't fix without extreme measures. 

The path of least resistance was to snapshot the DOM instead, which tbh I prefer.

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -109,6 +110,17 @@ export const OakModal = ({
...
+  // `createPortal` is not supported in SSR so we can only render when mounted on the client
+  const [isMounted, setIsMounted] = useState(false);
+
+  useEffect(() => {
+    setIsMounted(true);
+  }, []);
+
+  if (!isMounted) {
+    return null;
+  }
```
carlmw, line 122: tuesday

src/components/molecules/OakModal/OakModal.tsx
```diff
@@ -79,7 +80,7 @@ const logoSrc = `https://${process.env.NEXT_PUBLIC_OAK_ASSETS_HOST}/${process.en
 export const OakModal = ({
   children,
   footerSlot,
-  domContainer = document.body,
```
carlmw, line 82: This was the source of the error I was encountering locally — until now we hadn't tried the modal under SSR so this wasn't a problem. On the server `document` isn't defined so things break when this component is rendered.

simonrose121 (approved):
LGTM!

This is pretty how much Next's example handles the same issue: https://github.com/vercel/next.js/tree/canary/examples/with-portals

PR #208, opened 2024-06-24
chore: correct missing `OakGlobalStyle` import and add a note to the example

carlmw:
It was not abundantly clear that the example used the App router rather than Pages. We could also add a working example for the Pages router.

# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,11 @@ You can install it using `npm i @oaknational/oak-components` or any other packag
 
 For components to be styled correctly they will need access to a theme, some global styles and the Lexend font.
 
-You can add those to your app using something like:
-
-
-> 🚨 This example is for the [Next.js App router](https://nextjs.org/docs/app) . If you're using the [Pages router](https://nextjs.org/docs/pages/building-your-application) these changes should be made inside your `_document.js` 
-
+If you're using the Next.js App router your root layout should look something like:
 
 ```typescript
 // layout.js 
 import { OakThemeProvider, oakDefaultTheme, OakGlobalStyle } from "@oaknational/oak-components";
-import Head from "next/head";
 import { Lexend } from "next/font/google";
 
 const lexend = Lexend({ subsets: ['latin'] });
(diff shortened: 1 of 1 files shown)


remy:
In general this is fine, except for:

> these changes should be made inside your `_document.js`

This doesn't really work. It's a bit of a whack-a-mole, but first off Next complains that `next/font/google` can't be used in the _document. So I move that out, then I get errors, then the reset style isn't applied.

The only way I can get it to (nearly) work with pages setup is:

- _document.tsx is left alone (or customised, but doesn't include anything from this project)
- _app.tsx looks like this:

```tsx
import type { AppProps } from 'next/app';
import {
  OakThemeProvider,
  oakDefaultTheme,
  OakGlobalStyle,
} from '@oaknational/oak-components';
import { Lexend } from 'next/font/google';

const lexend = Lexend({ subsets: ['latin'] });

export default function OakOpenAPIApp({ Component }: AppProps) {
  return (
    <>
      <OakGlobalStyle />
      <main className={lexend.className}>
        <OakThemeProvider theme={oakDefaultTheme}>
          <Component />
        </OakThemeProvider>
      </main>
    </>
  );
}
```

Importantly that `OakGlobalStyle` is outside of the `main` element and _not_ in `Head` (which I really don't understand why).

But that all said, I still have hydration errors (which I'm sure I've seen with styled-components in the past, which is why my [_document.tsx](https://github.com/oaknational/oak-openapi/blob/2ec155860c81cee52ef4d67e9629ab193e221085/src/pages/_document.tsx) looks like this, but it still doesn't solve the hydration issue.

(nor does the font actually get used… I'm going to try to switch to app layout to see if all my problems go away…)

carlmw:
Okay, i've just tried it locally with my toy app using the "App" router and the `<Head>` is entirely redundant there too. I guess the "magic" of styled-components pulls it all into a `<style>` that it generates on the fly but this breaks under a pages router for some reason.

I've also added a note to say that Next needs to be configured for styled-components to do actual SSR.

I wish that these styles were compiled at build time instead of leaving it up to the host. Easier to reference them in a plain `<link>` but 🤷‍♂️

Sorry that this is all so complicated, but give me a shout if I can help rubber duck.

kimon-satan (approved):

PR #216, opened 2024-06-27
Add a note about needing env vars for icons

carlmw:
Icons just break at the moment. It might be worthwhile adding a check for the presence of the env vars when importing the library too.


kimon-satan (approved):
thanks :)

PR #232, opened 2024-07-17
[PUPIL-864] Fix accessibility issue on OakQuizHint

BTitterington:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

There was an issue with the `aria-describedby` referring to an id of an element that is only rendered when the button is clicked.

The fix is to place an additional hidden div with that id next to the component that contains the hint.

In addition, I noticed that with the `id` hardcoded it would mean we could only use the OakQuizHint once on a page. I adjusted it so the id is passed in. This also meant OakLessonBottomNav needed to include an `id` on the usage of OakQuizHint.

### DIFF
--- a/src/components/organisms/pupil/OakLessonBottomNav/OakLessonBottomNav.tsx
+++ b/src/components/organisms/pupil/OakLessonBottomNav/OakLessonBottomNav.tsx
@@ -41,7 +41,7 @@ export const OakLessonBottomNav = ({
       );
       break;
     case !!hint:
-      content = <OakQuizHint hint={hint} />;
+      content = <OakQuizHint hint={hint} id="quiz-hint" />;
       break;
     default:
       content = null;
--- a/src/components/organisms/pupil/OakQuizHint/OakQuizHint.tsx
+++ b/src/components/organisms/pupil/OakQuizHint/OakQuizHint.tsx
@@ -2,35 +2,37 @@ import React, { MouseEventHandler, ReactNode, useState } from "react";
 
 import { OakTooltip } from "@/components/molecules";
 import { OakHintButton } from "@/components/organisms/pupil/OakHintButton";
+import { OakBox } from "@/components/atoms";
 
 export type OakQuizHintProps = {
   /**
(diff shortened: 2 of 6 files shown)


src/components/organisms/pupil/OakQuizHint/OakQuizHint.tsx
```diff
@@ -2,35 +2,37 @@ import React, { MouseEventHandler, ReactNode, useState } from "react";
...
    */
   hint: ReactNode;
+  id: string;
 };
 
 /**
  * Presents a button which will show a hint when clicked
  */
-export const OakQuizHint = ({ hint }: OakQuizHintProps) => {
+export const OakQuizHint = ({ hint, id }: OakQuizHintProps) => {
```
kimon-satan, line 18: might be worth defaulting the id in any case ?

src/components/organisms/pupil/OakLessonBottomNav/OakLessonBottomNav.tsx
```diff
@@ -41,7 +41,7 @@ export const OakLessonBottomNav = ({
       );
       break;
     case !!hint:
-      content = <OakQuizHint hint={hint} />;
+      content = <OakQuizHint hint={hint} id="quiz-hint" />;
```
kimon-satan, line 44: A default would make this change unecessary

kimon-satan (approved):

src/components/organisms/pupil/OakQuizHint/OakQuizHint.tsx
BTitterington: I considered that, but then I'd have to make `id` optional and figured it is better to draw attention to the fact an `id` it is needed (if it ever gets used elsewhere). The other option was to use a generated guid for the id, but I figured I was overthinking it 🙂

RBrownsett (approved):
Approved - now passing accessibility  checker 
image

PR #250, opened 2024-07-30
Feat/new optionality card - LESQ-963

JBR90:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

-Adds new optionality card and optionality item component
-Adds grid-gap to grid component

## Link to the design doc

https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=9618-4307&m=dev

https://deploy-preview-250--lively-meringue-8ebd43.netlify.app/?path=/docs/components-organisms-teacher-oakunitlistoptionalityitem--docs


src/components/atoms/OakGrid/OakGrid.tsx
```diff
@@ -18,6 +22,7 @@ const gridStyle = css<OakGridProps>`
     (props) => props.$gridTemplateColumns,
   )}
     ${responsiveStyle("grid-template-rows", (props) => props.$gridTemplateRows)}
+    ${responsiveStyle("grid-gap", (props) => props.$gridGap, parseSpacing)}
```
carlmw, line 25: `grid-gap` [is a deprecated name](https://developer.mozilla.org/en-US/docs/Web/CSS/gap#:~:text=Early%20versions%20of%20the%20specification%20called%20this%20property%20grid%2Dgap).

We already use the correct names for `row-gap` and `column-gap` so we should probably use the unprefixed `gap` property here.

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+        {nullTitle}
+      </OakHeading>
+    </OakFlex>
+  );
+};
+
+export type OakUnitListOptionalityItemProps<C extends ElementType> = {
+  as?: C;
+  disabled?: boolean;
+  unavailable?: boolean;
```
njblore, line 115: do `disabled` and `unavailable` need to be two separate props? seems like they're always used in conjunction unless I'm missing it

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+            yearTitle={props.yearTitle}
+            optionalityUnitsLength={optionalityUnits.length}
+            disabledOrUnavailable={disabledOrUnavailable}
+          />
+        </OakFlex>
+        <OakGrid $gridGap={"space-between-xs"} role="list">
+          {optionalityUnits.map((unit, index) => (
+            <OakGridArea key={`${unit.title}-${index}`} $colSpan={[12, 6]}>
+              <OakUnitListOptionalityItemCard
+                key={unit.title}
```
carlmw, line 233: This `key` seems redundant

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
JBR90: They are always used in conjunction with each other, they both come from the extended `ElementType` type props rather than component , UnitListItem uses the same pattern so i kept it consistent.

JBR90: Maybe we don't need unavailable or either ?

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+              />
+            </OakFlex>
+          </OakFlex>
+        </OakFlex>
+        <OakBox $mb={"space-between-s"} $display={["none", "flex", "none"]}>
+          <StyledUnitHeading
+            nullTitle={nullTitle}
+            disabledOrUnavailable={disabledOrUnavailable}
+          />
+        </OakBox>
```
carlmw, line 216: Do we still want to leave the space for the margin when `nullTitle` is `undefined` and the element doesn't render?

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+      {index}
+    </OakSpan>
+  </OakFlex>
+);
+
+const StyledUnitHeading = ({
+  nullTitle,
+  disabledOrUnavailable,
+}: {
+  nullTitle: string | undefined;
```
carlmw, line 93: 🏄🏻 this is just my opinion (and it's an internal components so no biggy) but looking at LN106 `nullTitle` is the component's content so it should occupy the `children` slot.

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
njblore: Oh yeah I see, yeah it seems like we could just condense down to 1 or the other? save some confusion!

RBrownsett (commented):
@JBR90 getting an accessibility violation for contrast errors when the disabled state is shown (see screenshot). We had similar issues in pupil squad so we updated our disabled/unavailable font colour from rgb `128, 128, 128` to `87, 87, 87` - might be one for @Monika-Koziol to check and confirm if the designs need to be amended? 

image

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+    </OakBox>
+  </>
+);
+
+const StyledIndex = ({
+  index,
+  disabledOrUnavailable,
+  ...rest
+}: {
+  index: number;
```
carlmw, line 66: 🏄🏻 `children` ?

RBrownsett:
@JBR90 just checking - why is the legacy section show as unavailable? 

image

JBR90:
> @JBR90 just checking - why is the legacy section show as unavailable?
> 
@RBrownsett - sorry copy is misleading there, I will remove legacy. There are no legacy optionality units, this is meant to show unavailable units.

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+    lessonCount,
+    progress,
+    disabled,
+    href,
+    unavailable,
+    onClick,
+    index,
+    firstItemRef,
+    optionalityUnits,
+    nullTitle,
```
carlmw, line 151: A bunch of these props aren't mentioned in the type for this component's props,

Are you doing this so that you can omit them after spreading them from your consumer in OWA or is this an oversight?

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+        tag="h3"
+      >
+        {nullTitle}
+      </OakHeading>
+    </OakFlex>
+  );
+};
+
+export type OakUnitListOptionalityItemProps<C extends ElementType> = {
+  as?: C;
```
carlmw, line 113: This component doesn't appear to be polymorphic, do you need the `as`?

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
JBR90: I guess unavailable is if it is expired and disabled could be copyright / no logged in. But the style is the same atm so i say we stick with unavailable ?

src/components/organisms/teacher/OakUnitListOptionalityItem/OakUnitListOptionalityItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+    ...rest
+  } = props;
+
+  const disabledOrUnavailable = disabled || unavailable;
+
+  return (
+    <OakFlex
+      $flexDirection={["column", "row", "row"]}
+      $width={"100%"}
+      role="listitem"
```
carlmw, line 161: could this be `as="li"` instead of `role="listitem"`? I know you'd probably have to override the default list styling but the semantics are better and [Sonar will be happier](https://sonarcloud.io/project/issues?id=oaknational_oak-components&pullRequest=250&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

carlmw (commented):
🏄🏻 are just my reckons so you can ignore if you wish!

RBrownsett:
@JBR90 when there is an odd number of options (e.g. 3) does the third option align to the left of the component like it does currently?

Monika-Koziol:
@RBrownsett, just checking, do you mean the colour contrast is not accessible? Would it be the same rule as for buttons? Essentially, the colour contrast check does not apply as the component is disabled.

RBrownsett:
> @RBrownsett, just checking, do you mean the colour contrast is not accessible? Would it be the same rule as for buttons? Essentially, the colour contrast check does not apply as the component is disabled.

Yes sorry, the colour contrast for the text is not accessible - so it could difficult for users to read, for example those who have low vision - even though the content is disabled I would think that we would want the text to be legible so a user could see that it was disabled/still read the unit information.

Monika-Koziol (approved):

RBrownsett:
Agreed that font will remain the same as the designs based on https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html 

This may need to be revisited when the component is used in OWA and pa11y checks run against it

PR #329, opened 2024-11-21
CTC-56 Create OakMediaClipList and OakSolidBorderAccordion components

weronika-szalas:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

- added `OakMediaClipList` component (the list of media clip items on `/media` page) 
-  added `OakSolidBorderAccordion` component (because the styles of the accordion were different to any of the accordions we have on the site)

## Link to the design doc

https://www.figma.com/design/iTimVTBZ3XcszhRa9KtD7r/Cycle-2-requirements---Teachers?node-id=227-41162&node-type=frame&t=rjCSeCfNIKCIYjek-0

https://deploy-preview-329--lively-meringue-8ebd43.netlify.app/?path=/docs/components-organisms-teacher-oakmediacliplist--docs

### DIFF
--- a/src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
+++ b/src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
@@ -103,7 +103,7 @@ const Accordion = ({
           $mr={"space-between-xs"}
         />
       </StyledAccordionButton>
-      <InternalAccordionContent aria-labelledby={id}>
+      <InternalAccordionContent aria-labelledby={id} $overflow={"scroll"}>
         {children}
       </InternalAccordionContent>
       <StyledAccordionUnderline $fill={"border-decorative5-stronger"} />
@@ -115,12 +115,13 @@ const Accordion = ({
  * InternalChevronAccordion has a chevron icon that rotates when the accordion is open
  */
 
-export const InternalChevronAccordion = (
-  props: InternalChevronAccordionProps,
-) => {
+export const InternalChevronAccordion = ({
+  initialOpen = false,
+  ...props
+}: InternalChevronAccordionProps) => {
   return (
-    <InternalAccordionProvider isInitialOpen={props.initialOpen || false}>
-      <Accordion {...props} />
+    <InternalAccordionProvider isInitialOpen={initialOpen}>
+      <Accordion initialOpen {...props} />
     </InternalAccordionProvider>
   );
 };
--- a/src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
+++ b/src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
@@ -100,6 +100,7 @@ const Accordion = ({
             transform: isOpen ? "rotate(180deg)" : "none",
             transition: "all 0.3s ease 0s",
           }}
+          $mr={"space-between-xs"}
         />
       </StyledAccordionButton>
       <InternalAccordionContent aria-labelledby={id}>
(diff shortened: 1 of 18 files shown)


src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
```diff
@@ -100,6 +100,7 @@ const Accordion = ({
             transform: isOpen ? "rotate(180deg)" : "none",
             transition: "all 0.3s ease 0s",
           }}
+          $mr={"space-between-xs"}
```
weronika-szalas, line 103: this only affects `OakOutlineAccordion` where the change is barely visible so I decided to keep it here rather than add more props to the `InternalChevronAccordion`

src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
```diff
@@ -114,12 +115,13 @@ const Accordion = ({
...
 
-export const InternalChevronAccordion = (
-  props: InternalChevronAccordionProps,
-) => {
+export const InternalChevronAccordion = ({
+  initialOpen = false,
+  ...props
+}: InternalChevronAccordionProps) => {
   return (
-    <InternalAccordionProvider isInitialOpen={false}>
```
weronika-szalas, line 121: this was preventing accordion from being initially open

src/components/molecules/InternalChevronAccordion/InternalChevronAccordion.tsx
```diff
@@ -100,9 +100,10 @@ const Accordion = ({
             transform: isOpen ? "rotate(180deg)" : "none",
             transition: "all 0.3s ease 0s",
           }}
+          $mr={"space-between-xs"}
         />
       </StyledAccordionButton>
-      <InternalAccordionContent aria-labelledby={id}>
+      <InternalAccordionContent aria-labelledby={id} $overflow={"scroll"}>
```
weronika-szalas, line 106: Again, this only affects `OakOutlineAccordion` and only has effect if we set the `height` of accordion. The default was `overflow: visible` but I imagine with any overflow on accordion we want it to be `scroll` anyways

Monika-Koziol (changes requested):
@weronika-szalas coming along great. 
Screenshot 2024-11-21 at 13 35 06
- [ ] could we add some padding to the top and bottom of media clips? – inner-padding-ssx (4px)
Also more of a question: I've noticed we are using rems instead of design tokens. Just curious why?

weronika-szalas:
@Monika-Koziol Added `inner-padding-ssx` to the top of media clips (once you scroll is gone with the content so if we want it to stay we should use margin bottom on the header instead). Left out bottom padding as agreed in the chat that it looks weird with fade in effect at the bottom.

Not sure about rem question - we are using design tokens in the code.

RBrownsett:
@weronika-szalas @Monika-Koziol just on the above - are we removing the bottom fade? I think it should be removed so its clear when highlighted and matches the others

image

Monika-Koziol (approved):

RBrownsett:
> @RBrownsett @weronika-szalas, the bottom fade is meant to help users notice there are more clips in the box. I'm happy to remove it if others think it distracts rather than helps. @RBrownsett your call.

I like it as you scroll down @Monika-Koziol but it would be good that once you got to the bottom of list it stops fading. Is that possible to do @weronika-szalas ?

weronika-szalas:
@RBrownsett possible but will need more work to determine when the scroll is present and when is the end of the scroll, don't think we have this functionality anywhere in the app already, need to dig in more to see if it's a quick one.

RBrownsett:
> @RBrownsett possible but will need more work to determine when the scroll is present and when is the end of the scroll, don't think we have this functionality anywhere in the app already, need to dig in more to see if it's a quick one.

Yeah I just checking really, I don't want it to turn into a difficult/time consuming job.

weronika-szalas:
@RBrownsett  I would say let's get it in as it is if we want to merge today, otherwise I can look at it on Tuesday and either remove fade or play around detecting end of scroll to remove it.

k-huggs (approved):

PR #384, opened 2025-02-26
feat: add id const as exports

benprotheroe:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

exporting the ids as const so that they're is a single source of truth for the ids for the inputs


BTitterington:
Technically we shouldn't be using the same id on every element, they are supposed to be unique to one element on the page. Is there another way to do this where they get incremented or something like that?

benprotheroe:
> Technically we shouldn't be using the same id on every element, they are supposed to be unique to one element on the page. Is there another way to do this where they get incremented or something like that?

BTitterington (approved):

PR #421, opened 2025-05-20
Feat/lesq 1411/side menu nav [LESQ-1411]

njblore:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

- adds new shared organism `OakSideMenuNav`

## Link to the design doc
https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=14797-47586&p=f&m=dev

https://deploy-preview-421--lively-meringue-8ebd43.netlify.app/?path=/story/components-organisms-shared-oaksidemenunav--default

## Testing instructions
- I've added 2 stories, one regular length and one long list
- Check they match designs on desktop and mobile
- Check keyboard nav and a11y 

## ACs
In figma

### DIFF
--- a/src/components/organisms/shared/OakSideMenuNav/OakSideMenuNav.stories.tsx
+++ b/src/components/organisms/shared/OakSideMenuNav/OakSideMenuNav.stories.tsx
@@ -15,7 +15,6 @@ const meta: Meta<typeof OakSideMenuNav> = {
       },
     },
   },
-  args: {},
 };
 
 export default meta;
@@ -32,5 +31,27 @@ export const Default: Story = {
   ),
   args: {
     heading: "Test Heading",
+    menuItems: [
+      {
+        heading: "Test Item 1",
+        subheading: "Test Subheading 1",
+        href: "#test1",
+      },
+      {
+        heading: "Test Item 2",
+        subheading: "Test Subheading 2",
+        href: "#test2",
+      },
+      {
+        heading: "Test Item 3",
+        subheading: "Test Subheading 3",
+        href: "#test3",
+      },
+      {
+        heading: "Test Item 4",
+        subheading: "Test Subheading 4",
+        href: "#test4",
+      },
+    ],
   },
 };
--- a/src/components/organisms/shared/OakSideMenuNav/OakSideMenuNav.stories.tsx
+++ b/src/components/organisms/shared/OakSideMenuNav/OakSideMenuNav.stories.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { StoryObj, Meta } from "@storybook/react";
+
+import OakSideMenuNav from "./OakSideMenuNav";
+
+import { OakGrid, OakGridArea } from "@/components/atoms";
+
+const meta: Meta<typeof OakSideMenuNav> = {
+  component: OakSideMenuNav,
+  tags: ["autodocs"],
+  argTypes: {
+    heading: {
+      control: {
+        type: "text",
+      },
+    },
+  },
+  args: {},
+};
+
+export default meta;
+
+type Story = StoryObj<typeof OakSideMenuNav>;
+
+export const Default: Story = {
+  render: (args) => (
+    <OakGrid $background="pink" $width="100%" $height="100vh">
+      <OakGridArea $colSpan={[12, 2]}>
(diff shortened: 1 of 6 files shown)


left23 (approved):
This looks good!
There are a few aspects I am not sure about so maybe @Monika-Koziol can comment on these too.
1. On small mobile (from the breakpoint in the component library see https://deploy-preview-421--lively-meringue-8ebd43.netlify.app/?path=/story/components-organisms-shared-oaksidemenunav--default&globals=viewport:mobile1), is the heading and sub headings intended to wrap as they do? Or should they display one after the other on new lines,  and expand the remaining width, minus the chevron?
Selection_630

2. The width of the sidebar in the component library tablet view - see https://deploy-preview-421--lively-meringue-8ebd43.netlify.app/?path=/story/components-organisms-shared-oaksidemenunav--default&globals=viewport:tablet - is this the same width as in the designs or will it be once it's part of the page?  (notice the designs have a narrower width on tablet https://www.figma.com/design/LSjbzdvHo17XqXGBRNJjQP/Teacher-browse?node-id=8171-49781&p=f&m=dev)

3. I'm not sure if the navigation heading should be a h4 or not a heading at all... if I understand correctly, this might be set to a different heading level in the actual implementation? I am leaning towards it being a h2, to match the heading level of the Programme links in the main content area... I just wanted to point out that I am unsure so we might need to revisit this again once the page is built. I've also asked colleagues to share opinions on this (waiting to hear back)
None of these are blockers. I'm approving this so as not to hold up the implementation.
Selection_631

Taha-Hassan-Git (approved):
Seems good to me! I have a similar question to Lisa around the headings. If using an H4 should the nested elements use a heading element too? an h5? More of a question than a suggestion!

njblore:
Thanks @left23, on point 2 the width will be controlled by the page component in OWA, in the Story I've set it to take up 2 columns of a 12 column layout which matches the designs. And yeah for I wasn't too sure on the heading either, thinking about it again I'd probably go with it not being a heading, I'm not sure it's useful to have it as a heading as it's already in a nav anyway?

Monika-Koziol (changes requested):
@njblore looking great. A few small changes:

- [x] can the first item use the text-primary token? It's just slightly off-black we don't actually use pure black anymore, same for the border colour 
- [ ] it's a small detail but the black/green bar on the left has rounder corners. Not sure if that can be achieved with the border property. Not a biggie if it's a lot of work
- [ ] seeing it live it could do with a small gap between the black and grey text. Can you add 4px gap? My bad should've noticed that in the designs 
- [ ] for the mobile view: can the text wrap underneath each other rather than being in two columns? 
Comment rather than request but is this useful for naming HTML and accessibility https://www.w3.org/WAI/tutorials/menus/structure/

dkyyau (approved):

left23:
I think it depends on how we envision this component used in the future too. 

If there is always going to be a nav "title" element, using a h2 (or whatever heading level makes sense within the page's semantic context) is still appropriate.  Instead of using `aria-label` on the nav, we can have `aria-labelledby` with the heading ID as its value. So the heading would also need an ID in this case. The screenreader would read "Navigation - Collections" which is helpful.

If the design (or a future design) calls for a nav without a title element, but we wanted to stick with the pattern I mention above, the heading could still be present but visually hidden (with CSS) so it's still a good UX for screenreaders.

Otherwise, stick with the `aria-label` on the nav and this means it doesn't rely on the nav having a title element (visually hidden or not). This is probably the most flexible and then removes the need to determine which heading level to adopt depending on the implementation. In this case, I would do as you suggest @njblore and remove the heading from the title, and just use a div.

I hope that makes some sense. I don't have a preference.

left23 (approved):
Looks good, thanks for making those changes. Approving.

njblore:
@Monika-Koziol thanks I've resolved those issues except the rounded edges 💀 I don't think it's doable in the way I've implemented it as a border so would require a bit of a refactor - are we ok to leave it with sharp corners for now?

Monika-Koziol:
Thanks for looking into it @njblore, as I've said, not a biggie, most people wouldn't notice. I'll change in Figma so it matches.

PR #422, opened 2025-05-20
Feat/lesq 1415/save count [LESQ-1415]

njblore:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

- create new molecule `OakSaveCount`
- the prescribed width and height on the designs aren't valid spacing tokens in the component library so I used the closest available values

## Link to the design doc
https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=14263-4518&p=f&m=dev

https://deploy-preview-422--lively-meringue-8ebd43.netlify.app/?path=/docs/components-molecules-oaksavecount--docs

## ACs

- [ ] when loading the button should be greyed out with no hover or focus styles
- [ ] when not loading the styles should match the desins
- [ ] when the count is 0 the icon should be outlines
- [ ] when the count is > 0 the icon should be filled in
- [ ] when the count is greater than 99 the text should read '99+'

### DIFF
--- a/src/components/molecules/OakSaveCount/OakSaveCount.stories.tsx
+++ b/src/components/molecules/OakSaveCount/OakSaveCount.stories.tsx
@@ -18,6 +18,11 @@ const meta: Meta<typeof OakSaveCount> = {
         type: "text",
       },
     },
+    loading: {
+      control: {
+        type: "boolean",
+      },
+    },
   },
 };
 export default meta;
@@ -29,5 +34,6 @@ export const Default: Story = {
   args: {
     count: 1,
     href: "#",
+    loading: false,
   },
 };
--- a/src/components/molecules/OakSaveCount/OakSaveCount.stories.tsx
+++ b/src/components/molecules/OakSaveCount/OakSaveCount.stories.tsx
@@ -1,0 +1,33 @@
+import React from "react";
+import { Meta, StoryObj } from "@storybook/react";
+
+import { OakSaveCount } from "./OakSaveCount";
+
+const meta: Meta<typeof OakSaveCount> = {
+  component: OakSaveCount,
+  tags: ["autodocs"],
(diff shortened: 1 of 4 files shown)


left23 (changes requested):
This looks good. Nice idea with the 99+ - it removes the need for this to grow to fit more characters. Maybe the right edge of the background container is a bit close when we have 2+ characters.

It's great that the saved items count gets read out by the screenreader. I think we can go a bit further and add some text to help understand that the icon is also a link to the library page since there is no additional text to provide for the screenreader like there would be for normal hyperlinks.

I noticed there are a few components for visually-hidden elements - https://github.com/oaknational/Oak-Web-Application/blob/main/src/components/SharedComponents/ScreenReaderOnly/ScreenReaderOnly.tsx and https://github.com/oaknational/oak-components/blob/main/src/components/atoms/OakScreenReader/OakScreenReader.tsx
I'm not sure which is most relevant (they seem very similar) but it illustrates what I mean when I say visually hidden.

I would recommend something like:
```
<div>
<span aria-label="0 saved units" class="...">0</span>
<span class="srOnlyCss">View my library</span>
</div>
```
Where `srOnlyCss` is the class from whatever component you decide to use that visually hides the text.

That way we get the screenreader reading out "X saved units, view my library" which I think is nicely informative.

Pinging @Monika-Koziol here too in case there is a better line of copy than "View my library" - something more in line with language already in use?

Monika-Koziol:
"View my library' works fine in this place. @left23 
For the 99+ numbers might just increase the width ot the unit count. We might just have to adjust that on mobile when it's in the nav next to the sign-in button, as I get the feeling it won't all fit.

left23 (approved):
Thanks @njblore, approving.

PR #648, opened 2026-04-10
Fix: use unordered list on OakTabs to show grouping for screen readers

couplm:
# How to review this PR

- Check for these gotchas:
  - Circular dependencies
  - Relative imports
  - Default exports

- Replaced OakFlex with an unordered list to improve screen reader description of OakTabs.
- Grouped elements in an unordered list result in the following screen reader message, "visited, link, Description, 3 of 3" compared to previous message, "Download, button".
- Added a test that checks for the unordered list and the buttons in the list exist

## Link to the design doc

[Figma](https://www.figma.com/design/ktldjkEqAqv0X8PGzxrZyA/%F0%9F%9A%80-Integrated-journeys-v2?node-id=16876-63490&m=dev) - see "Programme Page v1 - Citizenship Secondary" specifically.

[Storybook preview](https://oak-components-storybook-git-fix-lesq-1853a11y-improvements.vercel.thenational.academy/?path=/docs/components-navigation-oaktabs--docs)

## Testing instructions

- Visit storybook preview above
- Navigate to OakTabs
- Turn on screen reader, I used mac voiceover
- See/hear screen reader message

## ACs

- [x]  links on programme page are grouped together with `<ul>` and `<li>` elements

### DIFF
--- a/src/components/navigation/OakTabs/OakTabs.tsx
+++ b/src/components/navigation/OakTabs/OakTabs.tsx
@@ -9,6 +9,7 @@ import { ResponsiveValues } from "@/styles/utils/responsiveStyle";
 import { InternalButton } from "@/components/internal-components/InternalButton";
 import { OakFlex } from "@/components/layout-and-structure";
 import { OakUiRoleToken } from "@/styles";
+import { OakUL } from "@/components/typography";
 
 type Tab<T> = {
   label: T;
@@ -32,7 +33,12 @@ export type OakTabsProps<T extends string> = {
 
 const StyledFocusOutline = styled(OakFlex)``;
 
-const StyledTabButton = styled(InternalButton)<{
+const StyledUnorderedList = styled(OakUL)`
+  display: flex;
+  justify-content: space-around;
+`;
+
+const StyledTabButton = styled(InternalButton) <{
   $hoverColor: OakUiRoleToken;
   $hoverBackground: OakUiRoleToken;
 }>`
@@ -61,7 +67,7 @@ export function OakTabs<T extends string>(props: Readonly<OakTabsProps<T>>) {
   const hoverText = colorVariant === "black" ? "text-inverted" : "text-primary";
 
   return (
-    <OakFlex
+    <StyledUnorderedList
       $height={sizeVariant === "compact" ? "spacing-40" : "spacing-56"}
       $background={backgroundColor}
       $borderRadius={"border-radius-circle"}
@@ -113,6 +119,6 @@ export function OakTabs<T extends string>(props: Readonly<OakTabsProps<T>>) {
           </StyledTabButton>
         );
       })}
-    </OakFlex>
+    </StyledUnorderedList>
   );
 }
--- a/src/components/navigation/OakTabs/OakTabs.tsx
+++ b/src/components/navigation/OakTabs/OakTabs.tsx
@@ -92,7 +92,7 @@ export function OakTabs<T extends string>(props: Readonly<OakTabsProps<T>>) {
             <StyledTabButton
               element={type === "link" ? Link : undefined}
               href={type === "link" ? tab.href : undefined}
-              aria-current={isSelected ? "page" : undefined}
+              aria-current={type === "link" && isSelected ? "page" : undefined}
               $background={isSelected ? "bg-decorative1-main" : backgroundColor}
               $color={isSelected ? "text-primary" : textColor}
               $hoverColor={isSelected ? "text-primary" : hoverText}
--- a/src/components/navigation/OakTabs/OakTabs.tsx
+++ b/src/components/navigation/OakTabs/OakTabs.tsx
@@ -33,7 +33,7 @@ export type OakTabsProps<T extends string> = {
 
 const StyledFocusOutline = styled(OakFlex)``;
 
-const StyledTabButton = styled(InternalButton) <{
+const StyledTabButton = styled(InternalButton)<{
   $hoverColor: OakUiRoleToken;
   $hoverBackground: OakUiRoleToken;
 }>`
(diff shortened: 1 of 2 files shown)


Taha-Hassan-Git (commented):
Looks good! We've got the unordered list element but the buttons also need to be wrapped in the list item component `<OakLI>`.

And from the ticket it seems like the whole thing needs to be wrapped in a navigation element, but maybe you're going to do that on the OWA side which would make sense.

src/components/navigation/OakTabs/OakTabs.tsx
```diff
@@ -61,7 +67,7 @@ export function OakTabs<T extends string>(props: Readonly<OakTabsProps<T>>) {
   const hoverText = colorVariant === "black" ? "text-inverted" : "text-primary";
 
   return (
-    <OakFlex
+    <StyledUnorderedList
```
njblore: now that this is a `ul` we'll need to wrap the buttons in `li` elements as well as lists should only contain `li`s ([see docs](https://www.w3.org/WAI/tutorials/page-structure/content/#unordered-list))

njblore: sorry just seen @Taha-Hassan-Git has already said that 😅 agree it probably makes sense to wrap in the `nav` in OWA as this won't always be used for nav (eg. when buttons)

couplm:
> Looks good! We've got the unordered list element but the buttons also need to be wrapped in the list item component `<OakLI>`.

Yeah, I agree completely about the list elements, I've just had so many problems trying to add them! Annoyingly I can't screenshot my voice over box, but the information becomes less useful and goes back to being "Download, button" when I add an li of any kind (Oak or not). I'll try and find a different way.

> And from the ticket it seems like the whole thing needs to be wrapped in a navigation element, but maybe you're going to do that on the OWA side which would make sense.

~~Yeah I wasn't entirely sure on use cases for OakTabs but maybe it does make more sense to put here if every OakTabs instance will be used like this?~~ Just seen what @njblore said, and so will keep it in OWA :)

couplm:
Tried using an li with the role presentation, didn't think it was ideal but it's the only way I can get it to show the same level of information...

Edit: [For reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role) I think it is fine to use in this use case, but maybe feels hacky

Taha-Hassan-Git:
> Tried using an li with the role presentation, didn't think it was ideal but it's the only way I can get it to show the same level of information...

I just tested and you're right that putting in the list items does kind of the opposite of what you might expect! The screen reader stops announcing where you are in the sequence. Not sure why, but maybe it's to avoid repeating info over and over. 

Personally I would have the list items, remove the presentation role and trust that the assistive technologies are handling it appropriately, it is a very common pattern after all.

EDIT: btw you can pass a prop to OakLI to set the list style to none, like this `<OakLI $listStyle={"none"}>`

couplm:
> I just tested and you're right that putting in the list items does kind of the opposite of what you might expect! The screen reader stops announcing where you are in the sequence. Not sure why, but maybe it's to avoid repeating info over and over.

Thanks! I feel validated knowing it's weird for you too and not me going crazy!

> Personally I would have the list items, remove the presentation role and trust that the assistive technologies are handling it appropriately, it is a very common pattern after all.

As we discussed just now, I'll defer to @RBrownsett when back on what to do regarding presentation role and the screen reader message :)

RBrownsett:
@couplm Nice work — I like that it now announces how many items are in the list 👍

Just need to fix the rendering of the link version. Also, for the link variant, does it programmatically indicate which item is the current page? (something I can double-check once it’s rendering properly).

We should ideally have something like:
aria-current={type === "link" && isSelected ? "page" : undefined}

couplm:
Thanks, @RBrownsett ! Just to clarify before I look at the link version, do you prefer it to be:

1. "Unit Sequence, button, (1 of 3), list 3 items", ... "Download, button, (3 of 3)"
2. "Unit Sequence, button, list 3 items", ... "Download, button"

For scenario 1, I could get it so styling is correct and the message has more info, but the caveat is adding the presentation role to Oak LI, which appears to be an [acceptable use](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role) of the role, but I can understand if you do not want to do this way.

Scenario 2 avoids it but is less informative and involves the current styling (which I can try and amend). Thanks again!

RBrownsett:
> Just to clarify before I look at the link version, do you prefer it to be:

Thanks — I think I’d prefer the second option if possible (announced as links rather than buttons) 

The native list semantics are useful. The more important thing for the link version is that the current item is exposed properly, e.g. with aria-current="page" on the active link.

So it should end up something like:

“Navigation region, Programme sections” (if wrapped in a nav with a label like Taha mentioned)
“List with 3 items”
“Unit sequence, link”
“Explainer, link”
“Download, link”

If one of them is the current page and you add aria-current="page", then we’d get something like:

“Unit sequence, link, current page”

src/components/navigation/OakTabs/OakTabs.tsx
```diff
@@ -79,40 +82,48 @@ export function OakTabs<T extends string>(props: Readonly<OakTabsProps<T>>) {
...
-              $height={"100%"}
-              $borderRadius={"border-radius-circle"}
-              $pa={"spacing-4"}
-              $alignItems={"center"}
-              $justifyContent={"center"}
-              $whiteSpace={"nowrap"}
+            <StyledTabButton
+              element={type === "link" ? Link : undefined}
+              href={type === "link" ? tab.href : undefined}
+              aria-current={isSelected ? "page" : undefined}
```
RBrownsett: Thinking about it we most probably only want it on the link version so could be updated to `aria-current={type === "link" && isSelected ? "page" : undefined} `but not a huge issue and happy to leave it as it is

RBrownsett (approved):
Nice work, happy to approve

PR #659, opened 2026-04-20
feat: unstyled chevron accordion can be both (un)controlled

stevegreenn:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

Enable the `UnstyledChevronAccordion` component to be both controlled/uncontrolled depending on incoming param usage.

## ACs

- [x] Accordion can be used both controlled/uncontrolled.

### DIFF
--- a/src/components/unstyled/UnstyledChevronAccordion/UnstyledChevronAccordion.tsx
+++ b/src/components/unstyled/UnstyledChevronAccordion/UnstyledChevronAccordion.tsx
@@ -46,17 +46,19 @@ type UnstyledChevronAccordionCommonProps = {
 type UnstyledChevronAccordionUncontrolledProps = {
   /** Whether the accordion should be open initially. Uncontrolled usage only. */
   isInitiallyOpen?: boolean;
+
   isOpen?: never;
   onOpenChange?: never;
 };
 
 type UnstyledChevronAccordionControlledProps = {
+  isInitiallyOpen?: never;
+
   /** Controlled open state. Must be paired with `onOpenChange`. */
   isOpen: boolean;
 
   /** Called when the user toggles the accordion. Required when `isOpen` is provided. */
   onOpenChange: (open: boolean) => void;
-  isInitiallyOpen?: never;
 };
 
 export type UnstyledChevronAccordionProps =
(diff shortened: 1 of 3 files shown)


src/components/unstyled/UnstyledChevronAccordion/UnstyledChevronAccordion.tsx
```diff
@@ -44,6 +43,29 @@ export type UnstyledChevronAccordionProps = {
...
+};
+
+type UnstyledChevronAccordionControlledProps = {
+  /** Controlled open state. Must be paired with `onOpenChange`. */
+  isOpen: boolean;
+
+  /** Called when the user toggles the accordion. Required when `isOpen` is provided. */
+  onOpenChange: (open: boolean) => void;
+  isInitiallyOpen?: never;
+};
```
AaronC81, line 62: Really slick to type-safely ensure you're using a component as either controlled or uncontrolled 👌

AaronC81 (approved):
LGTM, although might be worth also having a review from somebody more familiar with oak-components!

src/components/unstyled/UnstyledChevronAccordion/UnstyledChevronAccordion.tsx
jessicaab12, line 62: I think it reads nicer if the props are listed in the same order in both types. Also I don't love the name `onOpenChange` but I'm struggling to think of a better one - `onToggle` maybe?

jessicaab12 (changes requested):
Looks good! Just a small suggestion :)

src/components/unstyled/UnstyledChevronAccordion/UnstyledChevronAccordion.tsx
stevegreenn: Good call on the prop order. I'll fix that!

Re: the callback naming, `onOpenChange` is a fairly standard callback name/signature used across component libraries. It could be `onToggle(open: boolean) => void;` but I like `onOpenChange` since it's so explicit in behaviour.

jessicaab12: Ahh ok, makes sense!

jessicaab12 (approved):

PR #691, opened 2026-05-18
Feat/add scroll snap style

alexcupit:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

- Adds `scroll-margin-top` and `scroll-margin-left` to new `scrollSnapStyle` for `OakBox`

### DIFF
--- a/src/components/layout-and-structure/OakBox/OakBox.stories.tsx
+++ b/src/components/layout-and-structure/OakBox/OakBox.stories.tsx
@@ -14,30 +14,42 @@ import { borderArgTypes } from "@/storybook-helpers/borderStyleHelpers";
 import { opacityArgTypes } from "@/storybook-helpers/opacityStyleHelpers";
 import { zIndexArgTypes } from "@/storybook-helpers/zIndexStyleHelpers";
 import { transitionArgTypes } from "@/storybook-helpers/transitionStyleHelpers";
+import { scrollSnapArgTypes } from "@/storybook-helpers/scrollSnapStyleHelpers";
+
+const argTypes = {
+  ...colorArgTypes,
+  ...sizeArgTypes,
+  ...spacingArgTypes,
+  ...positionArgTypes,
+  ...borderArgTypes,
+  ...opacityArgTypes,
+  ...zIndexArgTypes,
+  ...transitionArgTypes,
+  ...scrollSnapArgTypes,
+};
 
 const meta: Meta<typeof OakBox> = {
   component: OakBox,
   tags: ["autodocs"],
   title: "components/Layout and structure/OakBox",
-  argTypes: {
-    ...colorArgTypes,
-    ...sizeArgTypes,
-    ...spacingArgTypes,
-    ...positionArgTypes,
-    ...borderArgTypes,
-    ...opacityArgTypes,
-    ...zIndexArgTypes,
-    ...transitionArgTypes,
-  },
+  argTypes,
   parameters: {
     controls: {
-      include: [],
+      include: Object.keys(argTypes),
     },
   },
 };
 export default meta;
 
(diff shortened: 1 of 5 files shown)


alexcupit:
Tempted to ignore the sonar cloud flag since I've followed the structure used in other types that would allow future expansion for other scroll snap properties, but no strong feelings if better to resolve the warning.

orangemug (approved):
Looks good to me, one thought should we be documenting these core props in the storybooks? Feels like another ticket if so but interested on folks opinions

alexcupit:
Good point, I've updated the OakBox default story to show all of the standard props - it was perhaps a bit misleading [before](https://components.thenational.academy/?path=/docs/components-layout-and-structure-oakbox--docs) that only `$background` and `$color` were available controls, but not sure if having the [full list](https://oak-components-storybook-git-feat-add-scroll-snap-style.vercel.thenational.academy/?path=/docs/components-layout-and-structure-oakbox--docs) is helpful or a bit noisy?  @CatMillerCM any thoughts on this?

CatMillerCM:
> Tempted to ignore the sonar cloud flag since I've followed the structure used in other types that would allow future expansion for other scroll snap properties, but no strong feelings if better to resolve the warning.

I'm not precious either way on this, but if you want to keep it expandable and sonar happy, I would be tempted to just name `ScrollMarginStyleProps` as `ScrollSnapStyleProps` from the get go, or yeah scrap the rename for now and just keep it as only controlling margin for the time being.

CatMillerCM:
> Good point, I've updated the OakBox default story to show all of the standard props - it was perhaps a bit misleading [before](https://components.thenational.academy/?path=/docs/components-layout-and-structure-oakbox--docs) that only `$background` and `$color` were available controls, but not sure if having the [full list](https://oak-components-storybook-git-feat-add-scroll-snap-style.vercel.thenational.academy/?path=/docs/components-layout-and-structure-oakbox--docs) is helpful or a bit noisy? @CatMillerCM any thoughts on this?

I agree that it's probably more misleading to only have some props shown.. in the standards we state to expose "Inherited props that are meaningful to the component, Ie, `$props`". This is pretty vague and open to interpretation of what is "meaningful", but as `OakBox` is meant to be extremely flexible, I'd say that most of these probs are actually quite crucial to its role in the library. `OakBox` and `OakFlex` I think are kind of outliers given that they are SO flexible and effectively wrappers in most cases, so will always accept so many more styles than a standard component by nature.

I feel like an ideal scenario to keep the exposed props but make it less noisy would be to only expose eg "borderRadius" as opposed to all of "btlr", "btrr", "bblr" and so on, and the same with margin, padding, border, position etc etc. Basically keeping the whole property which you can control the t/b/l/r independently through anyway. However, looking at the code and the way OakBox currently pulls in these props, it might be more effort than it's worth to separate these out and end up just meaning there's no source of truth as you're declaring different props in different places... 

Overall, I feel like `OakBox` isn't really something we'd be playing around with too much in Storybook anyway given its use cases, so I'm not *too* worried about it being super noisy, and do think it's helpful to see all it can inherit
Sorry for the essay 😄 What do you think? What's your gut saying? 😸

CatMillerCM (approved):
Replied to your comments but overall LGTM! Happy to get this through!

alexcupit:
Thanks for the comments! I've tidied up the type naming a bit to keep sonar happy and it might not need expanding in the future anyway. 

On the storybook side, my gut is it's actually quite good to have the props shown in their completeness. As someone quite new, storybook should be a useful source of docs and it's helpful being able to look there to see how a component can be controlled rather than having to navigate through the code. It's quite a long list but storybook does a lot to make it easy to navigate and skim through.

PR #694, opened 2026-05-20
Switch from npm to pnpm

orangemug:
## Description
Switch from `npm` to `pnpm` for package management which now includes

> Supply-chain protection on by default. [minimumReleaseAge](https://pnpm.io/settings#minimumreleaseage) defaults to 1440 (1 day) and [blockExoticSubdeps](https://pnpm.io/settings#blockexoticsubdeps) defaults to true.  

See <https://pnpm.io/blog/releases/11.0>

## RFC
[RFC: Switch to Pnpm for all repos](https://www.notion.so/oaknationalacademy/RFC-Switch-to-Pnpm-for-all-repos-27926cc4e1b18054b255e773720549a5)

## Testing instructions
Code review only

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,12 +29,12 @@ jobs:
         with:
           node-version: 24.x
       - name: Install dependencies
-        run: npm ci
+        run: pnpm ci
       - name: Verify the integrity of provenance attestations and registry signatures for installed dependencies
-        run: npm audit signatures
+        run: pnpm audit signatures
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
           HUSKY: 0
           CI: true
-        run: npx semantic-release
+        run: pnpx semantic-release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,12 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
+
+      - name: setup pnpm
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -24,17 +24,23 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
+
+      - name: setup pnpm
+        uses: pnpm/action-setup@08c4be7e2e672a47d11bd04269e27e5f3e8529cb #v6
+        with:
+          cache: true
+
       - name: Setup Node.js
```
CatMillerCM, line 33: I'm not sure, but do we also need `cache: "pnpm"` within this block?

package.json
```diff
@@ -24,7 +25,7 @@
     "build-storybook": "storybook build",
     "prepare": "husky install",
     "prepublishOnly": "./scripts/prepublish.sh",
-    "publish:local": "npm run build && yalc publish --no-scripts",
+    "publish:local": "pnpm build && yalc publish --no-scripts",
```
CatMillerCM, line 28: do we still need "run" here? so pnpm run ...?

orangemug: You can omit the `run` with `pnpm` so you can either `pnpm run build` or `pnpm build`

CatMillerCM: oh snazzy!

CatMillerCM (approved):

.github/workflows/release.yml
orangemug: Ta, fixed in https://github.com/oaknational/oak-components/pull/694/changes/c45c8a627188b23334b4ea00dd64896c071e40f8

PR #710, opened 2026-06-08
Fix/adopt 2038/oak link icon sizing

alexcupit:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

Removes `iconHeight` and `iconWidth` from `OakLink` props. Instead, the icon size is coupled to the font size. This also fixes alignment issues when a large icon is passed in.

Here are some examples where this would change:

## OWA before:
image
image

## OWA after:
image
image

## nc-alpha before:
image

## nc-alpha after:
image

https://oak-components-storybook-git-fix-adopt-2038oak-link-icon-sizing.vercel.thenational.academy/?path=/story/components-navigation-oaklink--with-font-prop

### DIFF
--- a/src/components/internal-components/InternalLink/InternalLink.tsx
+++ b/src/components/internal-components/InternalLink/InternalLink.tsx
@@ -48,7 +48,7 @@ const StyledLink = styled.a<{
   ${StyledOakIcon} {
     filter: ${(props) => parseColorFilter(props.$color)};
     display: inline-block;
-    vertical-align: bottom;
+    vertical-align: text-top;
   }
 
   &:focus-visible {
@@ -143,10 +143,7 @@ export const InternalLink: InternalLinkComponent = forwardRef(
       visitedColor,
       hoverColor,
       activeColor,
-      iconWidth = "spacing-24",
-      iconHeight = "spacing-24",
       textDecoration = "underline",
-      font,
       ...rest
     } = props;
 
@@ -166,8 +163,8 @@ export const InternalLink: InternalLinkComponent = forwardRef(
           return (
             <StyledOakIcon
               iconName={iconName}
-              iconWidth={iconWidth}
-              iconHeight={iconHeight}
+              iconWidth={"1.25em"}
+              iconHeight={"1.25em"}
             />
           );
         default:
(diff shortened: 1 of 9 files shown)


src/styles/theme/spacing.ts
```diff
@@ -84,6 +84,7 @@ export const additionalSpacingTokens = {
   inherit: "inherit",
   initial: "initial",
   unset: "unset",
+  "1.25em": "1.25em",
```
CatMillerCM: I'm not sure we should be adding this here. Ideally these values are all predefined by the design kit and follow a set pattern. I see why you're after `em` as it's relative to font size, but I think best to hardcode this into InternalLink as opposed to adding it to reusable styles. You won't then be able to pass it into `OakIcon`, but conveniently in this file we already have an empty `StyledOakIcon`. You could maybe use this like:
```
const StyledOakIcon = styled(OakIcon)`
  width: 1.25em;
  height: 1.25em;
`;
```

This would also mean InternalLink wouldn't need iconWidth or iconHeight props either which I think makes sense if we're removing support in OakLink

CatMillerCM (changes requested):

src/styles/theme/spacing.ts
alexcupit: Thanks Cat! Yeah I wasn't 100% happy with this bit and that's a nicer way of handling it! Will take a look!

HelenHarlow (commented):
One tiny thing from me, could we add 4px spacing between the link copy and the icon? It just looks a little tight, especially on the smallest text. Thanks!

alexcupit:
Thanks both! 
@CatMillerCM I've made those changes and also refactored the use of the `OakLoadingSpinner` as it was handled separately if you don't mind taking a look? 
@HelenHarlow I've added in that 4px between the link text and icon.

HelenHarlow (approved):
LGTM!

src/components/internal-components/InternalLink/InternalLink.tsx
```diff
@@ -14,11 +14,22 @@ import { OakBox } from "@/components/layout-and-structure/OakBox";
...
+  width: 1.25em;
+  min-width: 1.25em;
+  height: 1.25em;
+  min-height: 1.25em;
+`;
+
+const StyledOakLoadingSpinner = styled(OakLoadingSpinner)`
+  ${css`
+    --width: 1.25em;
+  `}
```
CatMillerCM: I think this might work without the css? so just
```suggestion
    --width: 1.25em;
```

CatMillerCM (approved):
Just one small comment, but otherwise LGTM!

PR #723, opened 2026-06-22
Refactor/oak download card

alexcupit:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

Refactors `OakDownloadCard` as per the new designs and new oak-component standards, specifically:
- adds `isEditable` prop and tag,
- accepts an array of formats and creates a `ul`,
- fixed styling: focus, hover, disabled and active states,
  - included an update to `OakDownloadsAccordion` to accommodate the focus box-shadow

## Link to the design doc

[Old](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=14795-4734&p=f&m=dev) and [New](https://www.figma.com/design/xIIH1NUk2IHKxgh5BW5gOd/Test-Oak-Component-Standards?node-id=66-905&p=f&m=dev)

[OakDownloadCard](https://oak-components-storybook-git-refactor-oak-download-card.vercel.thenational.academy/?path=/docs/components-owa-oakdownloadcard--docs)
[OakDownloadsAccordion](https://oak-components-storybook-git-refactor-oak-download-card.vercel.thenational.academy/?path=/docs/owa-teacher-oakdownloadsaccordion--docs)

### DIFF
--- a/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
+++ b/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
@@ -86,6 +86,12 @@ export type OakDownloadCardProps = BaseCheckBoxProps & {
    * @default false
    */
   isRadio?: boolean;
+  /**
+   * If true, adds an informational `Editable` tag to the download card.
+   *
+   * @default false
+   */
+  isEditable?: boolean;
 };
 
 const LabelContainer = styled("label")`
@@ -136,6 +142,7 @@ export const OakDownloadCard = (props: OakDownloadCardProps) => {
     onHovered,
     isRadio = false,
     "data-testid": dataTestId,
+    isEditable = false,
     ...rest
   } = props;
 
@@ -186,9 +193,14 @@ export const OakDownloadCard = (props: OakDownloadCardProps) => {
             $pv="spacing-12"
             $ph="spacing-16"
           >
-            <OakBox $font={"body-2-bold"}>{title}</OakBox>
-            {fileSize && <OakBox $font={"body-3"}>{fileSize}</OakBox>}
-            <OakBox $font={"body-3"}>{format}</OakBox>
+              {isEditable && (
+                <OakTagFunctional
+                  $alignSelf={"flex-end"}
+                  $font={"heading-light-7"}
+                  label="Editable"
+                  $background={"bg-decorative2-main"}
+                />
+              )}
           </OakFlex>
         </OakFlex>
         <OakFlex>
--- a/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
+++ b/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
@@ -75,9 +75,9 @@ export type OakDownloadCardProps = BaseCheckBoxProps & {
    */
   fileSize?: React.ReactNode;
   /**
-   * The file format or secondary metadata shown beneath the title.
+   * The file format(s) or secondary metadata shown beneath the title.
    */
-  format: React.ReactNode | React.ReactNode[];
+  format: string | string[];
   /**
    * The icon used to represent the download type.
    */
@@ -222,8 +222,8 @@ export const OakDownloadCard = (props: OakDownloadCardProps) => {
               <OakBox $font={"body-3"} $color={"text-subdued"}>
                 {Array.isArray(format) ? (
                   <OakUL $pl="spacing-20">
-                    {format.map((item, index) => (
-                      <OakLI key={index}>{item}</OakLI>
+                    {format.map((item) => (
+                      <OakLI key={item}>{item}</OakLI>
                     ))}
                   </OakUL>
                 ) : (
--- a/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
+++ b/src/components/owa/OakDownloadCard/OakDownloadCard.tsx
@@ -15,6 +15,8 @@ import { OakFlex } from "@/components/layout-and-structure/OakFlex";
 import { OakUiRoleToken } from "@/styles";
 import { parseColor } from "@/styles/helpers/parseColor";
 import { IconName } from "@/image-map";
+import { OakTagFunctional } from "@/components/messaging-and-feedback/OakTagFunctional";
+import { OakLI, OakUL } from "@/components/typography";
 
 const SingleIcon = ({ iconName }: { iconName: IconName }) => {
   return (
@@ -75,7 +77,7 @@ export type OakDownloadCardProps = BaseCheckBoxProps & {
   /**
    * The file format or secondary metadata shown beneath the title.
    */
-  format: React.ReactNode;
+  format: React.ReactNode | React.ReactNode[];
   /**
    * The icon used to represent the download type.
    */
@@ -193,6 +195,26 @@ export const OakDownloadCard = (props: OakDownloadCardProps) => {
             $pv="spacing-12"
             $ph="spacing-16"
           >
+            <OakBox id="download-card-title" $font={"body-2-bold"}>
+              {title}
+            </OakBox>
+            {fileSize && (
+              <OakBox $font={"body-3"} $color={"text-subdued"}>
+                {fileSize}
+              </OakBox>
+            )}
+            <OakFlex $alignItems={"center"} $gap={"spacing-8"}>
+              <OakBox $font={"body-3"} $color={"text-subdued"}>
+                {Array.isArray(format) ? (
+                  <OakUL $pl="spacing-20">
+                    {format.map((item, index) => (
+                      <OakLI key={index}>{item}</OakLI>
+                    ))}
+                  </OakUL>
+                ) : (
+                  format
+                )}
+              </OakBox>
               {isEditable && (
                 <OakTagFunctional
                   $alignSelf={"flex-end"}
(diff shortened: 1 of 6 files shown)


couplm:
I think this looks great 😄 

I noticed a few small things that don't matter for the functionality of this component, but writing them here just for clarity.

- With the disabled state toggled on in Storybook, there isn't a visual difference for the checkbox from the enabled state. It might be confusing if you can't click on it and see a visual clue. Probably one for design as it looks like from Figma disabled was agreed on after the designs were done :)
- In Storybook for the auto docs, clicking the checkbox doesn't change it to checked, and same for the radio outside of the radio group story. Looks like it was this way before your work though and likely just a side effect of autodocs.

src/components/owa/OakDownloadCard/OakDownloadCard.tsx
```diff
@@ -186,9 +210,35 @@ export const OakDownloadCard = (props: OakDownloadCardProps) => {
...
+              <OakBox $font={"body-3"} $color={"text-subdued"}>
+                {fileSize}
+              </OakBox>
+            )}
+            <OakFlex $alignItems={"center"} $gap={"spacing-8"}>
+              <OakBox $font={"body-3"} $color={"text-subdued"}>
+                {Array.isArray(format) ? (
+                  <OakUL $pl="spacing-20">
+                    {format.map((item, index) => (
+                      <OakLI key={index}>{item}</OakLI>
```
couplm: I guess we can presume that the item will be unique each time, can't think of a reason to have duplicate file formats shown. So we can swap out index perhaps?

freelancekhabib (approved):

src/components/owa/OakDownloadCard/OakDownloadCard.tsx
alexcupit: Yeah good spot! The format was previously typed as `ReactNode | ReactNode[]` and I was hesitant to use `key={format.toString()}` but I've also just updated the typing to `string | string[]` now since the only instance of passing in a node was to create the editable tag which is now handled by its own prop 👍

couplm (approved):

alexcupit:
Marking as draft since this was more experimental to try out the new oak component standards

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.tsx
rightsaidjames: This alt text describes the image and its purpose, but not its current state, and for screen reader users is made redundant by the `aria-expanded` prop on the button. I suggest one of the following:

* Empty alt text, which makes it a decorative image for assistive technology. 
  * I'm not sure about this solution, as it's not really a decorative image and the `aria-expanded` prop won't be exposed to all users who might need this info. But it's simple and would probably pass an audit.
* Give it alt text of 'Show all resources' when closed, and 'Hide all resources' when opened, then add `aria-hidden` to hide it from screen reader users and other potential AT that surfaces the `aria-expanded` value on the button.
  * I have tested this with VoiceOver and it seems to work well. The alt text describes the state of the icon, which could potentially be a useful fallback to a user who needs alt text but is accessing it using a tool other than a screen reader. Still, it's a sign that the whole button and its behaviour is overly complex and in need of a refactor.
* Give it alt text of 'Show all resources' when closed, and 'Hide all resources' when opened, and accept that screen reader users will effectively receive the same information twice.
  * This solution is less dodgy than the above one (there's no messing around with `aria-hidden` on a functional image that has alt text), but doesn't give quite as satisfying a result. Still, I think it's a safe and low-risk option.

rightsaidjames: Probably worth spinning this onto its own ticket, as it's pre-existing behaviour and might need several revisions to get right.

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.tsx
```diff
@@ -163,7 +163,7 @@ const Accordion = ({
         </OakFlex>
       </OakFlex>
       {/* Accordion content */}
-      <OakBox $position={"relative"} $overflow={"auto"}>
+      <OakBox $position={"relative"} $overflow={"visible"}>
         <InternalAccordionContent $pt={"spacing-20"} aria-labelledby={id}>
```
rightsaidjames:
```suggestion
        <InternalAccordionContent $pt={"spacing-20"} aria-label="Resource list">
```

The current accessible name for this region is too verbose and doesn't adequately describe the purpose/content of the group (`Select all resources Slides, quizzes, worksheet, additional materials An arrow to indicate whether the item is open or closed`). A simple aria-label of `Resource list` should do the trick here.

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.tsx
rightsaidjames: I think the underlying issue here is that the checkbox label is also the button text for the accordion, but the visible text doesn't actually make clear that there is an accordion that can be opened. The down/up chevron is a good indicator that there is an accordion available, but it doesn't have its own accessible name to go with it.

Therefore I think we should look into separating the checkbox label from the 'Show/hide all' button so we don't have to worry about maintaining visible text/accessible names that cover both controls.

alexcupit: Implemented the 3rd option from above for now

rightsaidjames: Looks good, and works as I expect in VoiceOver.

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.tsx
rightsaidjames: Confirmed as fixed - tested with VoiceOver.

src/components/internal-components/InternalAccordion/InternalAccordion.tsx
```diff
@@ -19,6 +19,10 @@ const FlexWithReset = styled(OakFlex)`
   padding: 0;
 `;
 
+type AccordionContentLabellingProps =
+  | { "aria-label": string; "aria-labelledby"?: never }
+  | { "aria-labelledby": string; "aria-label"?: never };
+
```
rightsaidjames, line 25: Just to confirm, does this code mean it can have `aria-label` or `aria-labelledby`, but not both?

src/components/owa/OakResourceCard/OakResourceCard.stories.tsx
```diff
@@ -1,75 +1,75 @@
...
 
+export const IsEditable: Story = {
+  render: (args) => <OakResourceCard {...args} />,
+  args: {
+    iconName: "book-steps",
+    id: "download-card-file-size",
+    value: "a test value",
+    title: "Fancy presentation",
+    description: "PPTX",
+    fileSize: "200KB",
```
rightsaidjames: For consistency with current OWA behaviour, can you modify the stories so there's a space between the number and the unit? e.g. `200 KB`, `1.4 MB`

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.stories.tsx
```diff
@@ -57,31 +57,31 @@ const WithState = (args: OakDownloadsAccordionProps) => {
...
       <OakFlex $flexDirection="column" $gap="spacing-16">
-        <OakDownloadCard
+        <OakResourceCard
           id="lesson-plan"
           value="lesson-plan"
           iconName="book-steps"
           title="Lesson plan"
-          format="PDF"
+          description="PDF"
           fileSize="245KB"
```
rightsaidjames: See above comment re: formatting of file size strings.

src/components/internal-components/InternalAccordion/InternalAccordion.tsx
alexcupit: Yep, and that either one of them is present. Previously `aria-labelledby` was mandatory and since we wanted to remove this and just use `aria-label` for the `OakResourcesAccordion`, I wanted to ensure that moving forward other uses of `InternalAccordion` would have either one of these present. It does also mean it can't have both but from my understanding `labelledby` takes priority so in effect, an element can't have both anyway?

rightsaidjames: In theory the accordion button wouldn't require either ARIA property if it had a visible text label (e.g. `Show/Hide`), but I'm guessing that this use case doesn't currently exist in the higher-level components.

And yes, `aria-labelledby` has higher priority than `aria-label`... so you could technically have both, but the `aria-label` would be redundant.

rightsaidjames (approved):
LGTM!

src/components/owa/teacher/OakResourcesAccordion/OakResourcesAccordion.stories.tsx
alexcupit: Fixed in [75c5549](https://github.com/oaknational/oak-components/pull/723/commits/75c554941c8b3ce778de9ce0066ce04a2581c4a6)

src/components/owa/OakResourceCard/OakResourceCard.stories.tsx
alexcupit: Fixed in [75c5549](https://github.com/oaknational/oak-components/pull/723/commits/75c554941c8b3ce778de9ce0066ce04a2581c4a6)

PR #728, opened 2026-06-25
fix: add responsive orientation to oak card

benprotheroe:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

- adds rsponsive design to oak card orientation
- adds background color for when using icons as the image

## Link to the design doc
Screenshot 2026-06-25 at 16 37 55

## Testing instructions
go to https://oak-components-storybook-git-fix-oak-card-responsivity.vercel.thenational.academy/?path=/story/components-navigation-oakcard--oak-subject-link-card

OakCard should match new design to replace old OakLinkCard

note that this card is in use in use in owa, checking that the current stories represent the use cases, and old cases are unchanged.


src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -80,16 +93,10 @@ export type OakCardProps = {
...
-  $pa: OakAllSpacingToken;
-  $gap: OakAllSpacingToken;
-};
+type StyledFlexProps = OakFlexProps;
 
 const StyledOakFlex = styled(OakFlex)<StyledFlexProps>`
-  flex-direction: ${({ $flexDirection }) => $flexDirection};
-  pa: ${({ $pa }) => parseSpacing($pa)};
-  gap: ${({ $gap }) => parseSpacing($gap)};
+  ${flexStyle}
```
benprotheroe, line 164: by using the OakFlex styles means we can pass responsive props into OakFlex, rather then redefining the props

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -108,20 +115,17 @@ const StyledOakFlex = styled(OakFlex)<StyledFlexProps>`
   }
 `;
 
-type StyledImageProps = {
-  $height: OakCombinedSpacingToken;
-  $width: OakCombinedSpacingToken;
-  $aspectRatio: "1/1" | "4/3";
+type StyledImageProps = SizeStyleProps & {
+  backgroundColor: OakUiRoleToken;
 };
```
benprotheroe: again here, height, width, aspect ratio are SizeStyleProps, so using this allows us to pass responsive props

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -149,7 +154,60 @@ export const OakCard = ({
...
+  const getImageHeight = (
+    orientation: "row" | "column",
+  ): OakCombinedSpacingToken => {
+    if (orientation === "row") {
+      return "spacing-240";
+    } else {
+      return "auto";
+    }
+  };
+
```
benprotheroe: these functions take the perscribed props form oakCard and match them to the prop token

src/components/navigation/OakCard/OakCard.stories.tsx
```diff
@@ -134,6 +142,32 @@ export const RowOrientationWithRectangularImage: Story = {
...
+    <OakBox>
+      <OakCard
+        heading="Check out our new digital literacy lessons!"
+        headingLevel="h1"
+        href={href}
+        cardWidth="100%"
+        imageSrc={generateOakIconURL("subject-digital-literacy")}
+        imageAlt="Illustration representing digital literacy"
+        subCopy="Lessons offering practical knowledge and skills, developed independently of the national curriculum."
+        subCopyColor="text-primary"
```
paulsweeting: Figma [design](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=23025-10094&m=dev) uses text-subdued (#575757) for the description.

Image

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -108,20 +115,17 @@ const StyledOakFlex = styled(OakFlex)<StyledFlexProps>`
...
 const StyledOakImage = styled(OakImage)<StyledImageProps>`
-  height: ${({ $height }) => parseSpacing($height)};
-  width: ${({ $width }) => $width};
-  aspect-ratio: ${({ $aspectRatio }) => $aspectRatio};
+  ${sizeStyle}
 
   img {
     border-radius: ${parseBorderRadius("border-radius-m2")};
     object-fit: cover;
+    background-color: ${(props) => parseColor(props.backgroundColor)};
```
paulsweeting: Minor difference here - Figma applies `bg-decorative1-main` to the 1:1 image container, not the image asset. 

This puts `background-color` on the img, which works for the current subject icon but may break if the image does not fill the 240×240 box. Consider using OakImage’s $background on the wrapper instead.

paulsweeting (changes requested):

src/components/navigation/OakCard/OakCard.tsx
benprotheroe: its my understanding that we create these global style props so that we dont have to redefine them in every component

paulsweeting (approved):

messmike (approved):

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -33,7 +38,15 @@ export type OakCardProps = {
...
    */
+  /**
+   * The background colour behind the image.
+   */
+  imageBackgroundColor?: OakUiRoleToken;
   cardOrientation?: "row" | "column";
+  /**
+   * The orientation of the card on small screens.
+   */
+  smallScreenOrientation?: "row" | "column";
```
charsville: This method of handling responsive design with an additional prop is quite different to how we have handled it in other oak-components. Normally we just use the array syntax and I think we can still do that here.

I think we just need a single cardOrientation but we make the type `ResponsiveValues<"row" | "column">`.

I will comment below how to make this work

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -149,7 +153,60 @@ export const OakCard = ({
   linkText,
   linkIconName = "arrow-right",
   hoverBackground = "bg-btn-secondary-hover",
+  imageBackgroundColor = "bg-neutral",
 }: OakCardProps) => {
+  const flexDirection = smallScreenOrientation
```
charsville: We have loads of conditionals in code here so that we can account for the 2 different props for responsiveness, we can replace a lot of this by using styled-components, but we will need to add a bit of logic for handling the Array approach to responsiveness. I will add another comment with code snippet on how to do that.

src/components/navigation/OakCard/OakCard.tsx
```diff
@@ -80,16 +93,10 @@ export type OakCardProps = {
   hoverBackground?: OakUiRoleToken;
 };
 
-type StyledFlexProps = {
-  $flexDirection: "row" | "column";
-  $pa: OakAllSpacingToken;
-  $gap: OakAllSpacingToken;
-};
+type StyledFlexProps = OakFlexProps;
```
charsville: Its around here with the StyledFlexProps that we can do all of the magic after changing the type of cardOrientation to a ResponsiveValue.

First we will want to create a new styled-component we can use to replace the OakFlex we render below the StyledOakImage. Its just so that it supports having css styles applied in a styled-components css block. 
So like this:
````ts
const CardContent = styled(OakFlex)``;
````

The next thing is that when you look at the styles in all of the conditionals in the code before, all it does is actually apply different styles depending on whether its a row or a column. So to simplify we can create a styled-components css block for each of them, and then use the CardContent above and the StyledOakImage inside the css block to style the child components.
e.g
````ts
const rowStyles = css`
  flex-direction: row;

  ${StyledOakImage} {
    height: ${parseSpacing("spacing-240")};
  }

  ${CardContent} {
    height: auto;
    padding-top: ${parseSpacing("spacing-24")};
    padding-bottom: ${parseSpacing("spacing-24")};
  }
`;

const columnStyles = css`
  flex-direction: column;

  ${StyledOakImage} {
    height: auto;
  }

  ${CardContent} {
    height: 100%;
    padding-top: ${parseSpacing("spacing-0")};
    padding-bottom: ${parseSpacing("spacing-0")};
  }
`;
````

The next bit is making sure we handle responsiveness, because now the types changed we might get an array where small screens its column, large screens row or vice-versa. 

We can do this with a helper that fetches the css block depending on row and column, and then another helper that checks if its not an array, and in that case just returns the relevant css style, but then if its an array, it will add the breakpoint code for small and large and render the [small, large] parts of the array css.

e.g:
````ts
const getCardOrientationStyles = (orientation?: "row" | "column" | null) =>
  orientation === "row" ? rowStyles : columnStyles;

const getResponsiveCardOrientationStyles = (
  cardOrientation: ResponsiveValues<"row" | "column">,
) => {
  if (!Array.isArray(cardOrientation)) {
    return getCardOrientationStyles(cardOrientation);
  }

  const [smallOrientation, largeOrientation] = cardOrientation;

  return css`
    ${getCardOrientationStyles(smallOrientation)};

    @media (min-width: ${getBreakpoint("small")}px) {
      ${getCardOrientationStyles(largeOrientation)}
    }
  `;
};
````

Final step is just passing the values through. Add the type to StyledOakFlex for passing the orientation value through and then call the helper:
````ts
type StyledFlexProps = OakFlexProps & {
  $cardOrientation: ResponsiveValues<"row" | "column">;
};
const StyledOakFlex = styled(OakFlex)<StyledFlexProps>`
  ${flexStyle}

  ${({ $cardOrientation }) =>
    getResponsiveCardOrientationStyles($cardOrientation)}
````

Now its way simpler to use when you render the React components because you can do:
````tsx
{/* Here we have removed flexDirection and replaced with the cardOrientation */}
<StyledOakFlex
        as="a"
        href={href}
        $cardOrientation={cardOrientation}
        $height={"100%"}
        $pa={"spacing-16"}
        $gap={"spacing-16"}
        $color={"text-primary"}
      >
...
{/* Further down where we use StyledOakImage we can remove the height becasue thats set in the css*/}
<StyledOakImage
            src={imageSrc || ""}
            alt={imageAlt || ""}
            $width={"auto"}
            $aspectRatio={aspectRatio}
            $background={imageBackgroundColor}
            $borderRadius={"border-radius-m2"}
          />
...
{/* Then where we had OakFlex replace that with CardContent and we can remove the padding and height because those get set by css block too */}
<CardContent
          $flexDirection="column"
          $justifyContent={"space-between"}
          $gap="spacing-16"
        >
````

In the storybook story when you use the component it now matches more how we would do responsive styling in other components like this:
```tsx
export const OakSubjectLinkCard: Story = {
  render: ({ href = "https://example.com" }) => (
    <OakBox>
      <OakCard
        heading="Check out our new digital literacy lessons!"
        headingLevel="h1"
        href={href}
        cardWidth="100%"
        imageSrc={generateOakIconURL("subject-digital-literacy")}
        imageAlt="Illustration representing digital literacy"
        subCopy="Lessons offering practical knowledge and skills, developed independently of the national curriculum."
        subCopyColor="text-subdued"
        tagBackground="bg-decorative1-main"
        linkText="Go to new digital literacy lessons"
        linkIconName="chevron-right"
        cardOrientation={["column", "row"]}
        imageBackgroundColor="bg-decorative1-main"
      />
    </OakBox>
  ),
  args: {
    href: "https://example.com",
  },
};
```

charsville (changes requested):

charsville (approved):

PR #740, opened 2026-07-16
Fix invalid weights for `body-*-bold` font tokens

orangemug:
# How to review this PR

_Leave this text block for the reviewer_

- Check [component hierarchy](https://miro.com/app/board/uXjVNnKBgyk=/?share_link_id=59445593794) is followed correctly
- Check the design [Heuristics](https://components.thenational.academy/?path=/docs/docs-howtodesigncomponents--docs#heuristics-for-component-design) have been followed
- Check [naming conventions](https://components.thenational.academy/?path=/docs/docs-namingconventions--docs) have been applied
- Check for these gotchyas:
  - Missing exports for Oak components
  - Accidental export of Internal components
  - Snapshots of unexpected components have been modified
  - Circular dependencies
  - Code duplication (via not using base components)
  - Non-functional storybook for this or related components
  - Sensitve files changed eg. atoms, or style tokens
  - Relative imports
  - Default exports

# Add your PR description below
Fix invalid weights for `body-*-bold` font tokens updating `body-*-bold` weight from `700` to `500`

## Link to the design doc
None

## A link to the component in the deployment preview
https://oak-components-storybook-git-fix-font-sizes-incorrect.vercel.thenational.academy/

## Testing instructions
Check font-weights are now correct

PR #745, opened 2026-07-23
Added `<OakVideo/>` component

orangemug:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

Added the new `<OakVideo/>` component

## Link to the design doc
https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=23577-2794&m=dev

https://oak-components-storybook-git-feat-adopt-2162-video-component.vercel.thenational.academy/?path=/docs/components-presentational-oakvideo--docs

### DIFF
--- a/src/components/presentational/OakVideo/OakVideo.stories.tsx
+++ b/src/components/presentational/OakVideo/OakVideo.stories.tsx
@@ -1,0 +1,99 @@
+import React from "react";
+import { StoryObj, Meta } from "@storybook/nextjs";
+
+import { OakVideo } from "./OakVideo";
+
+import { OakBox } from "@/components/layout-and-structure";
+
+const longText =
+  "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut et massa mi. Aliquam in hendrerit urna. Pellentesque sit amet sapien fringilla, mattis ligula consectetur, ultrices mauris. Maecenas vitae mattis tellus. Nullam quis imperdiet augue. Vestibulum auctor ornare leo, non suscipit magna interdum eu. Curabitur pellentesque nibh nibh, at maximus ante fermentum sit amet. Pellentesque commodo lacus at sodales sodales. Quisque sagittis orci ut diam condimentum, vel euismod erat placerat. In iaculis arcu eros, eget tempus orci facilisis id.Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut et massa mi. Aliquam in hendrerit urna. Pellentesque sit amet sapien fringilla, mattis ligula consectetur, ultrices mauris. Maecenas vitae mattis tellus. Nullam quis imperdiet augue. Vestibulum auctor ornare leo, non suscipit magna interdum eu. Curabitur pellentesque nibh nibh, at maximus ante fermentum sit amet. Pellentesque commodo lacus at sodales sodales. Quisque sagittis orci ut diam condimentum, vel euismod erat placerat. In iaculis arcu eros, Nullam quis imperdiet augue. Vestibulum auctor ornare leo, non suscipit magna interdum eu. Curabitur pellentesque nibh nibh, at maximus ante fermentum sit amet. Pellentesque commodo lacus at sodales ";
+
+const ARGS = {
+  videoSlot: (
+    <OakBox $aspectRatio={"30/17"} $background={"bg-neutral-stronger"}>
+      Placeholder
+    </OakBox>
+  ),
+  transcript: [longText, longText, longText],
+  heading:
+    "Building a research informed curriculum by adopting Oak’s foreign languages model",
+  body: "A couple of lines of supporting copy about the video to explain who is in the video and what they discuss.",
+  showTranscript: true,
+  showSignLanguage: true,
+  showCopyLink: true,
+};
+
+const meta: Meta<typeof OakVideo> = {
+  component: OakVideo,
+  title: "components/Presentational/OakVideo",
+  tags: ["autodocs"],
+  argTypes: {},
+  parameters: {},
+};
+
+export default meta;
+
+type Story = StoryObj<typeof OakVideo>;
+
+export const OnlyVideo: Story = {
+  render: (args) => {
+    return <OakVideo {...args} />;
+  },
+  args: {
+    videoSlot: ARGS.videoSlot,
+  },
+};
+
+export const WithHeadingAndBody: Story = {
+  render: (args) => {
+    return <OakVideo {...args} />;
+  },
+  args: {
+    videoSlot: ARGS.videoSlot,
+    heading: ARGS.heading,
+    body: ARGS.body,
+    showTranscript: ARGS.showTranscript,
+  },
+};
+
+export const WithTranscript: Story = {
+  render: (args) => {
+    return <OakVideo {...args} />;
+  },
+  args: {
+    videoSlot: ARGS.videoSlot,
+    heading: ARGS.heading,
+    body: ARGS.body,
+    transcript: ARGS.transcript,
+    showTranscript: ARGS.showTranscript,
+  },
+};
+
+export const WithSignLanguage: Story = {
+  render: (args) => {
+    return <OakVideo {...args} />;
+  },
+  args: {
+    videoSlot: ARGS.videoSlot,
+    heading: ARGS.heading,
+    body: ARGS.body,
+    transcript: ARGS.transcript,
+    showTranscript: ARGS.showTranscript,
+    showSignLanguage: ARGS.showSignLanguage,
+  },
+};
+
+export const WithCopyLink: Story = {
+  render: (args) => {
+    return <OakVideo {...args} />;
+  },
+  args: {
+    videoSlot: ARGS.videoSlot,
+    heading: ARGS.heading,
+    body: ARGS.body,
+    transcript: ARGS.transcript,
+    showTranscript: ARGS.showTranscript,
+    showSignLanguage: ARGS.showSignLanguage,
+    showCopyLink: ARGS.showCopyLink,
+  },
+};
--- a/src/components/presentational/OakVideo/OakVideo.stories.tsx
+++ b/src/components/presentational/OakVideo/OakVideo.stories.tsx
@@ -27,63 +27,41 @@ const meta: Meta<typeof OakVideo> = {
   component: OakVideo,
   title: "components/Presentational/OakVideo",
   tags: ["autodocs"],
-  argTypes: {},
-  parameters: {},
+  argTypes: {
+    heading: {
+      control: "text",
+    },
+    body: {
+      control: "text",
+    },
+    showTranscript: {
+      control: "boolean",
+    },
+    showSignLanguage: {
+      control: "boolean",
+    },
+    showCopyLink: {
+      control: "boolean",
+    },
+  },
+  parameters: {
+    controls: {
+      include: [
+        "heading",
+        "body",
+        "showTranscript",
+        "showSignLanguage",
+        "showCopyLink",
+      ],
+    },
+  },
 };
 
 export default meta;
 
 type Story = StoryObj<typeof OakVideo>;
 
-export const OnlyVideo: Story = {
-  render: (args) => {
-    return <OakVideo {...args} />;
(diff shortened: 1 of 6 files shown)


HelenHarlow (commented):
Looks good! A few small things I spotted:
• I think the video is missing the border, although you may have just not added it in yet
• In the designs, buttons should be using the small variant 
• The icons should be right rather than the left

orangemug:
Thanks @HelenHarlow fixed

- ✅ I think the video is missing the border, although you may have just not added it in yet
- ✅ In the designs, buttons should be using the small variant
- ✅ The icons should be right rather than the left

HelenHarlow (commented):
One more thing, is the text in the button using the right style? It looks like it's lighter than the components in the design kit

Image

Image

orangemug:
I think the design-kit might be wrong @HelenHarlow 

https://github.com/oaknational/oak-components/pull/740

Screenshot 2026-07-27 at 10 28 11
Screenshot 2026-07-27 at 10 28 00


`body-3-bold` is meant to be `500` weight

HelenHarlow:
> `body-3-bold` is meant to be `500` weight

how strange. I checked elsewhere on OWA, and they look like they do in the vid component, so I think it's all good. I'll update body-3-bold in the design kit

HelenHarlow (commented):
Sorry I did actually spot one more thing.. is it possible for the buttons to be reordered so that in the stacked layout the transcript button is directly above the transcript when opened?

Image


Image

orangemug:
@HelenHarlow fixed in https://github.com/oaknational/oak-components/pull/745/commits/aed08fb734e368ba680eb0f95afedfc470683311

HelenHarlow (approved):
LGTM!

CatMillerCM (commented):
I think there's still an extra gap when heading doesn't exist:
Image

src/components/presentational/OakVideo/OakVideo.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const ARGS = {
+  videoSlot: (
+    <OakBox $aspectRatio={"30/17"} $background={"bg-neutral-stronger"}>
+      Placeholder
+    </OakBox>
```
CatMillerCM: Is there a way to actually put a video in here, or does it get too annoying?

src/components/presentational/OakVideo/OakVideo.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const ARGS = {
```
CatMillerCM: not sure I'm a fan of this. If we're only using most of this in one place, can we just pass the headings etc into the main story, and then pass in the video placeholder twice to keep them decoupled and obvious?

src/components/presentational/OakVideo/OakVideo.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    expect(getByTestId("oak-video-transcript-container")).toHaveStyle(
+      "display: none",
+    );
+    const showTranscriptButton = getByRole("button", {
+      name: "Show transcript",
+    });
+    await user.click(showTranscriptButton);
+    expect(getByTestId("oak-video-transcript-container")).not.toHaveStyle(
+      "display: none",
+    );
```
CatMillerCM: here, can we just test for `TEST_1`, `TEST_2` etc to be/not be in the document? Ideally would avoid testing on style and with testId if can be avoided

src/components/presentational/OakVideo/OakVideo.stories.tsx
orangemug: I think video is an OWA concern, I mean I could but an empty video element in there but it wouldn't really show anything off because Mux is the underlying player

orangemug:
> I think there's still an extra gap when heading doesn't exist:

@CatMillerCM fixed in latest, also added missing `headingTag`

src/components/presentational/OakVideo/OakVideo.stories.tsx
orangemug: Fixed in https://github.com/oaknational/oak-components/pull/745/commits/e0b426d20d91e67e69d310b4dadcadba3249a67b

src/components/presentational/OakVideo/OakVideo.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+  transcript: [longText, longText, longText],
+  heading:
+    "Building a research informed curriculum by adopting Oak’s foreign languages model",
+  body: "A couple of lines of supporting copy about the video to explain who is in the video and what they discuss.",
+  showTranscript: true,
+  showSignLanguage: true,
+  showCopyLink: true,
+};
+
+const meta: Meta<typeof OakVideo> = {
```
CatMillerCM, line 11: would it be worth also exposing `transcript` for testing do you think?

src/components/presentational/OakVideo/OakVideo.test.tsx
orangemug: It'll always be in the document because it's just hidden, I've changed it to `.not.toBeVisible()` & `.toBeVisible()` better?

src/components/presentational/OakVideo/OakVideo.tsx
```diff
@@ -0,0 +1,139 @@
...
+        )}
+        {body && (
+          <OakP $color={"text-primary"} $font={["body-2", "body-1", "body-1"]}>
+            {body}
+          </OakP>
+        )}
+      </OakFlex>
+      {hasButtonsEnabled && (
+        <OakFlex
+          $flexDirection={["row", "row", "row"]}
```
CatMillerCM:
```suggestion
          $flexDirection={"row"}
```

CatMillerCM: or is this your preference to always display as an array even if the same?

orangemug: Nah just had to change some stuff around half way and didn't notice they had all become the same, fixed in latest

src/components/presentational/OakVideo/OakVideo.tsx
```diff
@@ -0,0 +1,146 @@
...
+                onClick={onShowSignLanguage}
+                iconName={"sign-language"}
+              >
+                Show sign language
+              </OakSmallSecondaryButton>
+            </OakFlex>
+          )}
+        </OakFlex>
+      )}
+      {transcriptEnabled && (
```
CatMillerCM, line 153: apologies if I'm missing something here, but can this not just be driven off `isTranscriptOpen` so that the transcript is only rendered on click of show transcript (which can only be clicked if showTranscript is true, transcript exists and transcript > 0 etc)?

orangemug: The browser has to do more work and parse on "click" that's a perf thing. Also `id={transcriptId}` would be missing from the DOM, so I assume a11y tools wouldn't like that with `aria-controls={transcriptId}` in the button above

CatMillerCM: I guess this way the transcript (if it exists) is always visible in the DOM right, whether a user has chosen to show it or not? Is that preferable for screen readers etc? Or do screen readers skip if something has display none?
cc @rightsaidjames maybe

src/components/presentational/OakVideo/OakVideo.tsx
```diff
@@ -0,0 +1,146 @@
+import React, { useState, useId } from "react";
+
+import type { OakHeadingProps } from "@/components/typography";
+import { OakSmallSecondaryButton } from "@/components/buttons";
+import { OakFlex, OakBox } from "@/components/layout-and-structure";
+import { OakHeading, OakP } from "@/components/typography";
+
+export type OakVideoProps = {
+  headingTag?: OakHeadingProps["tag"];
```
CatMillerCM, line 14: there is a slight mismatch between storybook and the allowed values here (div, undefined). Should we align them? Should the video allow "div" to be passed for the heading tag?

src/components/presentational/OakVideo/OakVideo.tsx
```diff
@@ -0,0 +1,146 @@
+import React, { useState, useId } from "react";
+
+import type { OakHeadingProps } from "@/components/typography";
+import { OakSmallSecondaryButton } from "@/components/buttons";
+import { OakFlex, OakBox } from "@/components/layout-and-structure";
+import { OakHeading, OakP } from "@/components/typography";
+
+export type OakVideoProps = {
```
CatMillerCM, line 8: sorry to be annoying, but could we please document according to https://app.notion.com/p/oaknationalacademy/Oak-Components-Standards-v3-35726cc4e1b180f58ccef8cc58274b90?source=copy_link#35726cc4e1b180a58d6ece8ab9c61c88
(copilot come through)

src/components/presentational/OakVideo/OakVideo.tsx
```diff
@@ -0,0 +1,181 @@
...
+   */
+  showTranscript?: boolean;
+  /**
+   * Whether to show the sign language button.
+   */
+  showCopyLink?: boolean;
+  /**
+   * Whether to show the copy link button.
+   */
+  showSignLanguage?: boolean;
```
rightsaidjames: These comments were the wrong way around
```suggestion
   * Whether to show the copy link button.
   */
  showCopyLink?: boolean;
  /**
   * Whether to show the sign language button.
   */
  showSignLanguage?: boolean;
```

Image

src/components/presentational/OakVideo/OakVideo.tsx
rightsaidjames: @CatMillerCM Screen readers generally won't read out content that is `display:none`'d, but the fact that there is a semantic relationship between the button and the transcript div (via `aria-controls`) means that (IMO) it's good practice to keep it in the DOM when not shown. Also there might be a SEO benefit to rendering the transcript on page load rather than on demand.

rightsaidjames: Side note: `aria-controls` is [not well supported by most screen readers](https://a11ysupport.io/tech/aria/aria-controls_attribute) (JAWS being the exception), but that doesn't mean it shouldn't be used where appropriate.

rightsaidjames (approved):
Assuming the above switcheroo with the docs is fixed, LGTM!

src/components/presentational/OakVideo/OakVideo.tsx
orangemug: Yeah probably, will update now

orangemug: Fixed in https://github.com/oaknational/oak-components/pull/745/commits/bada03b89be4869e8642397fd751726e6f3140c1 `undefined` doesn't change the behaviour so opted with leaving that, also it's inline with what we do elsewhere

We should probably also fix `<OakCard/>` in the same way, but I'll leave that for another ticket.

src/components/presentational/OakVideo/OakVideo.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+    await user.click(showSignLanguageButton);
+    expect(onShowSignLanguage).toHaveBeenCalled();
+
+    expect(getByTestId("oak-video-transcript-container")).not.toBeVisible();
+    const showTranscriptButton = getByRole("button", {
+      name: "Show transcript",
+    });
+    await user.click(showTranscriptButton);
+    expect(getByTestId("oak-video-transcript-container")).toBeVisible();
```
CatMillerCM, line 75: real nit: I would be tempted to move these into a new test just as I feel like they're separate enough from the standard render of elements so that if something were to start failing I could isolate it more easily

CatMillerCM (approved):
LGTM! Happy to get this merged (party parrot)

src/components/presentational/OakVideo/OakVideo.test.tsx
orangemug: Fixed

PR #747, opened 2026-07-24
Feat/con 1972 update oakbuttonwithdropdown

jessicaab12:
# How to review this PR

- Check for these gotchyas:
  - Circular dependencies
  - Relative imports
  - Default exports

#

This PR updates `OakButtonWithDropdown` so that it can be used to build `MultiSelectDropdown` in Studio. It allows a new `close` function to be passed to the component's `children` so that it can handle when the dropdown is closed.

Added some new props that are optional so there should be no impact on existing implementations

Tidied up the tests and stories while I'm here for `OakButtonWithDropdown` and changed story names from camelCase to PascalCase where flagged!

Also features a tweak to `package.json` so that use/remove-local-components works in other repos

## Link to the design doc
[Design](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=20551-718&m=dev&focus-id=20620-6014)

Note that this design is for the `MultiSelectDropdown` component that will be built in Studio and this PR simply facilitates that so incorrect spacing, for example, can be ignored here 

[Preview](https://oak-components-storybook-git-feat-con-1972-update-oakbut-39029c.vercel.thenational.academy/?path=/story/components-buttons-oakbuttonwithdropdown--close-handled-by-children)

## Testing instructions
Check existing stories work as expected
On [this story](https://oak-components-storybook-git-feat-con-1972-update-oakbut-39029c.vercel.thenational.academy/?path=/story/components-buttons-oakbuttonwithdropdown--close-handled-by-children) the dropdown should permit multiple items to be selected and close when `Close` is clicked


AaronC81 (approved):
LGTM - great stuff tidying up some of the tests and stories too 🙌 the simplified `closeOnChange` tests are ace

I think the "`children` is a node or a function" pattern is new for oak-components. I've seen it before in other React projects and I think it works well here, but might be good to get a second pair of eyes from somebody who's more experienced with oak-components than me on that! Wonder if it's also worth including the pattern in the standards somewhere.

Passing data down to children is done with context in some rare cases ([accordions](https://github.com/oaknational/oak-components/blob/c16c3fdff5300c3ce781410cba8d091c45fe70ae/src/components/internal-components/InternalAccordion/InternalAccordionProvider.tsx#L9) and [radio button groups](https://github.com/oaknational/oak-components/blob/c16c3fdff5300c3ce781410cba8d091c45fe70ae/src/components/form-elements/OakRadioGroup/OakRadioGroup.tsx#L21)) but these are quite internal so not really the same use-case.

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx
```diff
@@ -199,12 +209,15 @@ export const OakButtonWithDropdown = ({
...
+            role={dropdownRole}
+            aria-label={
+              dropdownAriaLabel ??
+              "Dropdown menu. Use arrow keys to navigate, Tab to cycle through items, Escape to close."
+            }
             data-testid={dataTestId ? `${dataTestId}-dropdown` : undefined}
           >
             <OakFlex $flexDirection="column" $gap={"spacing-8"}>
-              {children}
+              {typeof children === "function" ? children(close) : children}
```
orangemug: I think the preferred is to use react context over function props. This is an old react method from prior to hooks

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx
```diff
@@ -38,6 +38,12 @@ export type OakButtonWithDropdownProps = {
   >;
   flexWidth?: ResponsiveValues<OakCombinedSpacingToken | null | undefined>;
   closeOnChange?: boolean;
+
+  /** The ARIA role for the dropdown panel. Defaults to "menu". */
+  dropdownRole?: React.AriaRole;
```
orangemug: The norm here is passing as `aria-*`, how about

```js
{
 dropdownProps: {
   "aria-role": "..."
   "aria-label": "..."
 } 
}
``` 

So we don't end up with a mix between `aria-role` and `*ariaRole` across the API?

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx
orangemug: So instead of

```tsx
<OakButtonWithDropdown>
  {close => <>
    <button onClick={close}>Close</button>
  </>}
</OakButtonWithDropdown>
```

This instead

```tsx
function Inner () {
  const {onClose} = useDropDownProps();
  
  return (
    <button onClick={onClose}>close</button>
  )
}

<OakButtonWithDropdown>
  <Inner/>
</OakButtonWithDropdown>
```

Also has `onClose` rather than `close` with is not obviously an event handler.

orangemug: The problem as I see it, is that the `children` are always rendering in your approach when the higher component renders. Which you probably want to avoid, otherwise you can easily end up in a situation where you're rendering all-the-things on change for unrelated prop changes

jessicaab12: Will update to use context instead, thanks!

> The problem as I see it, is that the children are always rendering in your approach when the higher component renders.

Not sure I follow this part though 🤔 Why is it the case that the children not always rendering in the second approach?

orangemug: @jessicaab12 this does a pretty good job of explaining it https://www.developerway.com/posts/react-elements-children-parents

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx
```diff
@@ -38,6 +40,12 @@ export type OakButtonWithDropdownProps = {
   >;
   flexWidth?: ResponsiveValues<OakCombinedSpacingToken | null | undefined>;
   closeOnChange?: boolean;
+  dropdownProps?: {
+    /** The ARIA role for the dropdown panel. Defaults to "menu". */
+    "aria-role"?: React.AriaRole;
```
stevegreenn: I think this should remain as `role`, `aria-role` is not a real attribute so this is a bit confusing IMO.

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.stories.tsx
```diff
@@ -198,3 +203,30 @@ export const ButtonWithDropdownCloseOnChange: OakButtonWithDropdownStory = {
     </OakFlex>
   ),
 };
+
+export const CloseHandledByChildren: OakButtonWithDropdownStory = {
+  decorators: [withReactContext],
```
stevegreenn: Do we need this? I think we can remove:

```tsx
  decorators: [withReactContext],
  parameters: {
    reactContext: {
      context: dropdownContext,
      contextValue: { close: () => fn() },
    },
  },
```

and then both the imports, and then remove the export from `dropdownContext`.

`OakButtonWithDropdownProvider` is already inside `OakButtonWithDropdown` so we don't to wrap the story with the context here.

stevegreenn (changes requested):

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.stories.tsx
jessicaab12: Yeah that was left over from something I tried before! Removed now

stevegreenn (approved):
bosh

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.test.tsx
```diff
@@ -24,6 +26,18 @@ const simpleChildren = (
   </>
 );
 
+const ChildrenUsingContext = () => {
+  const { close } = useDropdownContext();
```
orangemug: The react norm is `on*` for example https://reactuse.com/state/usedisclosure/

So this would be `onClose`

Does that match the standards @CatMillerCM ?

orangemug (changes requested):

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdownProvider.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const OakButtonWithDropdownProvider = ({
+  close,
+  children,
+}: {
+  close: () => void;
+  children: React.ReactNode;
+}) => {
+  const context: DropdownContextType = {
+    close,
+  };
```
orangemug: Good catch sonarcloud, 

```ts
const context = useMemo(() => {close}, [close]);
```

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdownProvider.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { createContext } from "react";
+
+type DropdownContextType = {
+  close: () => void;
+};
+
+export const dropdownContext = createContext<DropdownContextType | null>(null);
```
orangemug: `dropdownContext` -> `DropdownContext`

orangemug (changes requested):

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.test.tsx
CatMillerCM: yeah I would agree with onClose if passing behaviour (not a bool).
I actually think we'd want to update this section of the standards to include this
image

orangemug:
Few changes, otherwise looks good

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.test.tsx
jessicaab12: `close` simply closes the dropdown with no additional behaviour which is why I opted for that name. I felt `onClose` might be misleading here? @orangemug @CatMillerCM

CatMillerCM: Sorry, I think I need to get more context here - I'll have a look in the morning if that's okay? Brain is mush now 😄 
But have noticed we have `close` and `closeOnChange`, do they equate?

orangemug: I personally like the pattern because you know what it is and you know that it can be passed to an event handler. The `on*` methods like `onClose` signify a couple of things to the developer

It can be passed directly to a handler, so you wouldn't expect a argument to every be added to a `on*` handler later on. For example if you pass `onClose` directly to a `<button onClick={onClose} />` then you later extend the API. It's be odd to do anything other than add a single argument `(e: Event) => {}`

This a common approach which protect us from breaking changes in the future. In your example you're not signifying it can be directly passed, so you'd expect it's usage to be `<button onClick={() => close()} />` and not `<button onClick={close} />` it's because the `on*` comes with some norms across the react ecosystem.

It's also the name the consumer would most expect with a React app, so this is (should) be the pattern

```jsx
const onAction = () => {}
<button onClick={onAction} />
```

```jsx
const action = () => {}
<button onClick={action} />
```

Essentially `onSomething`'s are designed to go on things like `onClick`/`onBlur`

orangemug: FYI: Happy to be convinced otherwise, just want us to be consistent :)

jessicaab12: Yeah that's alright! It's blocking a Studio PR so keen to get it over the line in the morning if poss!

The changes here allow the dropdown to be closed by children. Our [planned use case](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=20551-718) in Studio is this `MultiSelectDropdown` component where multiple checkboxes can be selected, then a button clicked to close the dropdown.

`closeOnChange` closes when one item is clicked so we can't use it here

jessicaab12: That makes sense, thanks @orangemug! Updated!

orangemug (approved):

paulsweeting:
@jessicaab12 Ah -- Looks like we have an accessibility issue. When OakButtonWithDropdown closes (via onClose() from useDropdownContext, Escape, click outside, or closeOnChange), focus is not moved back to the trigger button. If focus was on a child inside the panel (e.g. Close/Add button or a menu item), that element unmounts and focus is lost — especially noticeable with a screen reader.

Although the component never did this. This PR adds useDropdownContext / onClose, which makes it more visible. Looking at where this component was already used in OWA, it's never been an issue before as the dropdown remains open. Is this something fixable in this PR? 

**Steps To Replicate:** 
- [Preview Env](https://oak-components-storybook-git-feat-con-1972-update-oakbut-39029c.vercel.thenational.academy/?path=/story/components-buttons-oakbuttonwithdropdown--close-handled-by-children)

- Using the Tab key, highlight the dropdown button
- Press space to trigger the dropdown
- Use the arrow keys or tab key to highlight the 'Close' button
- Press space / Enter to close the dropdown 

**Expected:**
Focus is returned to the dropdown button 'Trigger'

**Actual**
Focus is lost

**Video Demonstration:** 

https://github.com/user-attachments/assets/86774eb5-da99-443f-a5ae-ce109f01886d

paulsweeting:
@jessicaab12  dropdownProps lets you override the panel role and aria-label, but the trigger is still hardcoded as a menu (aria-haspopup="menu" — [line 191](https://github.com/oaknational/oak-components/blob/feat/con-1972-update-oakbuttonwithdropdown/src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx#L191)).


axe flags role="menu" when children are checkboxes + a button. `menu` isn’t the right role for the multi-select pattern, we need a different panel role (or disclosure without aria-haspopup)

https://dequeuniversity.com/rules/axe/4.12/aria-required-children?application=AxeChrome

Screenshot 2026-07-30 at 10 06 41

---------------
--- Update -- 

@jessicaab12 - As a suggested different approach to using 'menu', please see this [PR](https://github.com/oaknational/oak-components/pull/751) and [Link to Preview env](https://oak-components-storybook-git-feat-con-1972-disclosure-dr-2d2636.vercel.thenational.academy/?path=/story/components-buttons-oakbuttonwithdropdown--close-handled-by-children) 

Added dropdownType "disclosure" for multi-select

src/components/buttons/OakButtonWithDropdown/OakButtonWithDropdown.tsx
CatMillerCM, line 1: could we please [add descriptions](https://app.notion.com/p/oaknationalacademy/Oak-Components-Standards-v3-35726cc4e1b180f58ccef8cc58274b90?source=copy_link#35726cc4e1b1809198f3c34ed4fd82ca) for all the props? Might help clarify also
Also, I think maybe we could expose more of the props in storybook?

jessicaab12: Definitely worth doing but I think it may need to be done in a later PR as I really need to get this merged asap and a dependent Studio reviewed today sorry!

CatMillerCM (approved):

paulsweeting (approved):
Previous Accessibility issues have now been resolved. 👍

